Issue #8, opened 2019-11-07
Compilation errors for unix port on macOS


mdaeron:
I am able to compile micropython+ulab for the PYBD_SF6 but not for unix.

My batch script is the following:

```
mkdir build
cd build

git clone https://github.com/micropython/micropython.git
git clone https://github.com/v923z/micropython-ulab.git ulab

cd micropython
git submodule update --init

cd mpy-cross/
make

cd ../ports/stm32
echo '#define MODULE_ULAB_ENABLED (1)' | cat >> mpconfigport.h
make BOARD=PYBD_SF6 CROSS_COMPILE=/usr/local/bin/arm-none-eabi- USER_C_MODULES=../../../ulab all

cd ../unix
echo '#define MODULE_ULAB_ENABLED (1)' | cat >> mpconfigport.h
make USER_C_MODULES=../../../ulab all
```

When running the above code, I get a functional PUBD_SF6 firmware but the unix make results in 3 `-Wparentheses-equality` errors in `ulab/code/ndarray.c`:

```
CC ../../../ulab/code/ndarray.c
../../../ulab/code/ndarray.c:583:20: error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
    if((ndarray->m == 1)) {
        ~~~~~~~~~~~^~~~
../../../ulab/code/ndarray.c:583:20: note: remove extraneous parentheses around the comparison to silence this warning
    if((ndarray->m == 1)) {
       ~           ^   ~
../../../ulab/code/ndarray.c:583:20: note: use '=' to turn this equality comparison into an assignment
    if((ndarray->m == 1)) {
                   ^~
                   =
../../../ulab/code/ndarray.c:891:39: error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
            if((self->array->typecode == NDARRAY_INT8)) {
                ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
../../../ulab/code/ndarray.c:891:39: note: remove extraneous parentheses around the comparison to silence this warning
            if((self->array->typecode == NDARRAY_INT8)) {
               ~                      ^              ~
../../../ulab/code/ndarray.c:891:39: note: use '=' to turn this equality comparison into an assignment
            if((self->array->typecode == NDARRAY_INT8)) {
                                      ^~
                                      =
[...]

PR #12, opened 2019-11-11
Fixed several issues preventing compilation on macOS

nickovs:
Using the original code, attempting to compile on macOS using `clang` would fail. There were three types of error that were raised:

 * In `ndarray.c` there were three occurrences of the condition of `if` statements being an equality comparison surrounded by double parentheses. This raises a warning (treated as an error due to compilation with `-Werror`) because use of double parentheses is a standard idiom for _suppressing_ the warning about use of an assignment in a condition. This was fixed by removing the excess parentheses.
 * In `linalg.c` there were three cases of `abs()` being called with a floating point value and the result being compared to a floating point value, even though `abs()` takes and returns `int` values. This was replaced with calls to `fabs` as used elsewhere in the same file.
 * In `poly.c` static control flow analysis complained about potential use of un-initialised variable (`deg` and `leny`). While manual analysis indicates that there is no path that allows the variables to be used in an uninitialised state, simply initialising the variables with 0 removed the static analysis issue without increasing code size.

With these three issues fixed the code compiles cleanly on macOS using `clang`.

### DIFF
--- a/code/linalg.c
+++ b/code/linalg.c
@@ -124,7 +124,7 @@ bool linalg_invert_matrix(mp_float_t *data, size_t N) {
     }
     for(size_t m=0; m < N; m++){
         // this could be faster with ((c < epsilon) && (c > -epsilon))
-        if(abs(data[m*(N+1)]) < epsilon) {
+        if(fabs(data[m*(N+1)]) < epsilon) {
             m_del(mp_float_t, unit, N*N);
             return false;
         }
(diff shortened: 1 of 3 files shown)


v923z:
Thanks! I believe, this issue is identical to https://github.com/v923z/micropython-ulab/issues/8, and has been fixed in the testing branch. Is that correct? I can merge it into master, if you want.

v923z:
Nicko,

>     * In `linalg.c` there were three cases of `abs()` being called with a floating point value and the result being compared to a floating point value, even though `abs()` takes and returns `int` values. This was replaced with calls to `fabs` as used elsewhere in the same file.

Shouldn't this be MICROPY_FLOAT_C_FUN(fabs) ?
https://github.com/v923z/micropython-ulab/blob/50df00546b2e7a172daff7100bc0382caa7b1629/code/linalg.c#L374

Otherwise, the code won't compile for all platforms. 

Unfortunately, I had missed this before merging your commint.

PR #22, opened 2019-12-31
preliminary edit to work with CircuitPython

GathererA:


dhalbert:
If you have stylistic things like `MP_OBJ_IS_TYPE` -> `mp_obj_is_type`, you can propose changes to CircuitPython too, to align us with MicroPython. We need to make these changes eventually anyway.

code/linalg.c
```diff
@@ -274,16 +274,16 @@ mp_obj_t linalg_eye(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args)
...
-    if((k >= 0) && (k < n)) {
-        while(k < n) {
+    if((k >= 0) && (abs(k) < n)) {
+        while(abs(k) < n) {
             mp_binary_set_val_array(dtype, ndarray->array->items, i*n+k, one);
             k++;
             i++;
         }
-    } else if((k < 0) && (-k < m)) {
+    } else if((k < 0) && (abs(k) < m)) {
```
v923z: One should use the macro MICROPY_FLOAT_C_FUN(fabs), I believe, but might be OK.

code/linalg.c
```diff
@@ -79,13 +79,13 @@ mp_obj_t linalg_size(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args)
     mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
     mp_arg_parse_all(1, pos_args, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
-    if(!mp_obj_is_type(args[0].u_obj, &ulab_ndarray_type)) {
-        mp_raise_TypeError("size is defined for ndarrays only");
+    if(!MP_OBJ_IS_TYPE(args[0].u_obj, &ulab_ndarray_type)) {
```
v923z: By using the macro, we add to the code side, though, the penalty is probably small in this case.

code/linalg.c
```diff
@@ -274,16 +274,16 @@ mp_obj_t linalg_eye(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args)
     ndarray_obj_t *ndarray = create_new_ndarray(m, n, dtype);
     mp_obj_t one = mp_obj_new_int(1);
     size_t i = 0;
-    if((k >= 0) && (k < n)) {
-        while(k < n) {
+    if((k >= 0) && (abs(k) < n)) {
```
v923z: If k >= 0, then abs(k) = k, isn't it? I don't see what the absolute value is supposed to do here...

code/linalg.c
```diff
@@ -274,16 +274,16 @@ mp_obj_t linalg_eye(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args)
...
             mp_binary_set_val_array(dtype, ndarray->array->items, i*n+k, one);
             k++;
             i++;
         }
-    } else if((k < 0) && (-k < m)) {
+    } else if((k < 0) && (abs(k) < m)) {
         k = -k;
         i = 0;
-        while(k < m) {
+        while(abs(k) < m) {
```
v923z: Is the absolute value necessary here?

code/ndarray.c
```diff
@@ -78,7 +78,7 @@ void ndarray_print_row(const mp_print_t *print, mp_obj_array_t *data, size_t n0,
         }
         mp_printf(print, ", ..., ");
         mp_obj_print_helper(print, mp_binary_get_val_array(data->typecode, data->items, n0+n-3), PRINT_REPR);
-        for(size_t i=1; i<3; i++) {
+        for(i=1; i<3; i++) {
```
v923z: Must the declaration of `i` really go? The old version made `i` local to that particular loop, which I find easier to work with.

code/ndarray.c
```diff
@@ -164,16 +164,19 @@ STATIC uint8_t ndarray_init_helper(size_t n_args, const mp_obj_t *pos_args, mp_m
     return dtype;
 }
 
-mp_obj_t ndarray_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
-    mp_arg_check_num(n_args, n_kw, 1, 2, true);
-    mp_map_t kw_args;
-    mp_map_init_fixed_table(&kw_args, n_kw, args + n_args);
-    uint8_t dtype = ndarray_init_helper(n_args, args, &kw_args);
+mp_obj_t ndarray_make_new(const mp_obj_type_t *type, size_t n_args, const mp_obj_t *args, mp_map_t *kw_args) {
```
v923z: I think this won't compile under micropython, so we really need a switching mechanism for this case.

v923z:
> If you have stylistic things like `MP_OBJ_IS_TYPE` -> `mp_obj_is_type`, you can propose changes to CircuitPython too, to align us with MicroPython. We need to make these changes eventually anyway.

That is not simply a stylistic issue. Some types will work only with the macro, I believe. I don't recall right away, which types were the offending ones. I chose the function, because that produces smaller code (mp_obj_is_type is already compiled into micropython, while MP_OBJ_IS_TYPE will be expanded each time), but I don't have any preferences at this point.

code/linalg.c
dhalbert: This is probably to fix a signed vs unsigned comparison warning, since CircuitPython has `-Wall` set. Currently `n` is not validated as > 0, which it probably should be. After validation, an intermediate `validated_n` that is a `uint32_t` or `size_t` could be used instead.

code/ndarray.c
dhalbert: The compiler is probably warning that the `size_t i=1;` on line 81 is shadowing `size_t i;` line 66. The line 66 declaration could be removed and the `for` loops could each have their own declaration.

code/ndarray.c
dhalbert: Yes, we (CircuitPython) changed the way `...make_new()` is called to accommodate something (I think it may have been better support of Python subclasses of native classes). Damien was supportive of this but felt he could not change the current MicroPython internal API due to so many modules depending on it.

Note that we have a preprocessor value that can be tested to determine if it's CircuitPython.
```c
#if CIRCUITPY
...
```
Also note we in general try to use `#if` tests rather than `#ifdef` because they're safer, and we also usually turn on the warning that will complain if `FOO` is undefined for
`#if FOO`

dhalbert:
> > If you have stylistic things like `MP_OBJ_IS_TYPE` -> `mp_obj_is_type`, you can propose changes to CircuitPython too, to align us with MicroPython. We need to make these changes eventually anyway.
> 
> That is not simply a stylistic issue. Some types will work only with the macro, I believe. I don't recall right away, which types were the offending ones. I chose the function, because that produces smaller code (mp_obj_is_type is already compiled into micropython, while MP_OBJ_IS_TYPE will be expanded each time), but I don't have any preferences at this point.

`mp_obj_is_type()` is a macro now:
https://github.com/micropython/micropython/blob/10709846f38f8f6519dee27694ce583926a00cb9/py/obj.h#L628

code/ndarray.c
v923z: This is actually my bad, thanks for pointing this out. You are right, line 66 should just be removed. Actually, in the long run it won't be needed at all: https://github.com/v923z/micropython-ulab/blob/2867ec9f13024167fcb22fb06a2e46b1e7fea46d/code/ndarray.c#L76

code/ndarray.c
v923z: OK, I can go with that, no problem.

v923z:
> `mp_obj_is_type()` is a macro now:
> https://github.com/micropython/micropython/blob/10709846f38f8f6519dee27694ce583926a00cb9/py/obj.h#L628

That's right, so we should just used that everywhere. I assume, it was changed in v.1.12. You mentioned that you run with 1.9.4. If using `mp_obj_is_type()` cases any problems in circuitpython, as a temporary fix, we have to define that as a substitute for `MP_OBJ_IS_TYPE` in a 
```
#if CIRCUITPY
...
```
pre-processor statement. 

I don't know what @GathererA's standpoint is. He outlined the required changes. I can implement those, and see if we can compile the code for circuitpython and micropythnon. But he mentioned some issues related to run-time imports, and I don't have the means of resolving that, so I would need some input from your side.

code/linalg.c
GathererA: yeah its a signed, unsigned comparison so I figured this was an easy fix?

code/linalg.c
GathererA: is there any issue with using abs()?

code/linalg.c
GathererA: nope, I'll take it out

code/linalg.c
dhalbert: Using `abs()` is tricking things up, when the true fix is to make sure `n` is >=  zero and then cast it appropriately to an unsigned. As I mentioned, the code should be validating `n`, since a negative value will cause havoc, I think.

code/ndarray.c
GathererA: cool, good to know, I'll switch it around in my fork

GathererA:
Sorry for the confusion, this is my first time editing code for CircuitPython or MicroPython. Currently the code compiles, but I'm still working through how to actually add the module to CircuitPython. I've forked CircuitPython and will be adding the ulab to the shared modules list for automatic inclusion, and then I can go back once I've got it working and outline the differences?

code/linalg.c
GathererA: since n is unsigned, wouldn't we want to create an intermediate validated_k that is unsigned if k>=0?

code/linalg.c
v923z: OK. (Might be a bit expensive with the abs function, but I don't think it is a deal-breaker.)

code/linalg.c
v923z: I am not against it, I was just wondering what was going on.

v923z:
> Sorry for the confusion, this is my first time editing code for CircuitPython or MicroPython. Currently the code compiles, but I'm still working through how to actually add the module to CircuitPython. 

To so how you can include a module, you could take something really simple from the usermod link that I mentioned earlier. 

> I've forked CircuitPython and will be adding the ulab to the shared modules list for automatic inclusion, and then I can go back once I've got it working and outline the differences?

Sounds good. You can fetch the circuitpython branch of ulab, though, I haven't yet pushed any changes.

code/linalg.c
v923z: Fine with me.

dhalbert: Well, I guess, both `n` and `k` need validation, since the code does:
```c
    size_t n = args[0].u_int, m;
    int16_t k = args[2].u_int;
```
if `n` is negative, that will cause it to be a very large positive value. And `k` could be negative, but `int16_t` could also end up truncating `k`. I'm not familiar with the API to know what the restrictions on these args are. @v923z, are you avoiding arg validation for speed reasons? I imagine there are other places where there is code like this.

code/ndarray.c
v923z: Before you are carrier away, that is the ndim branch:) For now, we should just yank line 66, and add `size_t i=` in the for loops.

code/linalg.c
v923z: I don't think that we should be concerned with speed here: what you suggest is a single `if` statement, or something similar, while the code is meant to iterate over arrays, and apply whatever mathematical functions to them. That is gonna be much more expensive than just validating the argument at the beginning. So, if we need to do that, then let's do that.

PR #29, opened 2020-02-05
convolve: implement and document

jepler:


jepler:
Testing performed: built the Unix port and ran the example shown.  Manually verified against numpy output for consistency.

v923z:
Brilliant, many thanks Jeff! A couple of small comments: 

1. could you, please, change the version number at at beginning of ulab.c to 0.27.0?
2. you should also modify the changelog in docs/ . You can add your name there, I don't mind. 
3. if you want to, you can remove a lot of the includes at the beginning of filter.c. I think, obj.h, runtime.h and builtin.h are enough, but I can't test at the very moment.

Otherwise, this is a nice addition, thanks again for contributing!

v923z:
@jepler Jeff, I changed a couple of small things in filter.c, fac3eb4, namely, I bail out, if the input array is not linear, and I replaced the ternary operator by if clauses, because they are approx. 25 bytes cheaper each.

jepler:
Thanks!

PR #31, opened 2020-02-08
ndarray: let memoryview(arr) work

jepler:
.. this makes ndarray.rawbytes redundant.  Circuitpython will remove it.

### DIFF
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -910,3 +910,9 @@ mp_obj_t ndarray_unary_op(mp_unary_op_t op, mp_obj_t self_in) {
         default: return MP_OBJ_NULL; // operator not supported
     }
 }
+
+mp_int_t ndarray_get_buffer(mp_obj_t self_in, mp_buffer_info_t *bufinfo, mp_uint_t flags) {
+    ndarray_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    // buffer_p.get_buffer() returns zero for success, while mp_get_buffer returns true for success
+    return !mp_get_buffer(self->array, bufinfo, flags);
+}
--- a/code/ndarray.h
+++ b/code/ndarray.h
@@ -69,6 +69,8 @@ mp_obj_t ndarray_rawsize(mp_obj_t );
 mp_obj_t ndarray_flatten(size_t , const mp_obj_t *, mp_map_t *);
 mp_obj_t ndarray_asbytearray(mp_obj_t );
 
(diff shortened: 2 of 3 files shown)


v923z:
> .. this makes ndarray.rawbytes redundant.

Did you mean `asbytearray` https://github.com/v923z/micropython-ulab/blob/918075daaf5f68c2145aca567a8036f67f9829af/code/ndarray.c#L682 ? 

I totally agree with the PR. However, we should then decide, what happens, if you have a sliced array as in https://github.com/v923z/micropython-ulab/blob/d240e5aad28be84da0b74974074fd6645868a0f3/code/ndarray.c#L184 The memory location in this case is not necessarily contiguous, because only the `strides` member is set. What should we return in that case? A dense copy, perhaps?

jepler:
As far as I saw, `memoryview(arr.asbytearray())` and the new `memoryview(arr)` would behave the same.  Are the limitations you mention, of ndarrays with nonzero `stride`, a problem in both?

If there are cases where it can't work, I think circuitpython would raise an exception.

v923z:
> As far as I saw, `memoryview(arr.asbytearray())` and the new `memoryview(arr)` would behave the same. Are the limitations you mention, of ndarrays with nonzero `stride`, a problem in both?

Yes, as soon as the 1dim/ndim branches are merged into master. The current master returns a copy anyway, so there it shouldn't be a problem.

> 
> If there are cases where it can't work, I think circuitpython would raise an exception.

OK, let's do that. I will merge this, and put a TODO flag in the function. I also remove the `asbytearray` function, and update the manual.

jepler:
Will it always return a copy, or just in the case of slicing? In circuitpython we want to use this to dma data from dac to ndarrays. Incidentally that's a reason to need int16 as a data type if you are keeping a list

v923z:
> Will it always return a copy, or just in the case of slicing? 

At the moment, it will always return a copy, but not for long. I am really close to wrapping up a proper implementation that works with views. Give me a couple of days.

> In circuitpython we want to use this to dma data from ADC to ndarrays. Incidentally that's a reason to need int16 as a data type if you are keeping a list

We could/should expose the `.data` attribute: https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.ndarray.data.html#numpy.ndarray.data, which lays bare the bytes. 

I assume, it doesn't matter, whether it is int16, or uint16. You just want to have words, right?

PR #35, opened 2020-02-11
convolve: Optimize and special-case floats

jepler:
Special casing floats decreases runtime to about 45% of original (applying a 117-tap filter to 512 points of data goes from 70ms to 32ms)

The top_n/bot_n calculations already meant that the a/c indices were never out of range.  This decreases runtime further to about 15% of original (11ms)

Timings done on an Adafruit Clue (nrf52840 at 64MHz)

It does of course increase code size somewhat.  It would be possible to put the all-float case inside an `ifdef` if that's a tradeoff you want to offer to users of ulab.

### DIFF
--- a/code/filter.c
+++ b/code/filter.c
@@ -47,23 +47,41 @@ mp_obj_t filter_convolve(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_a
     ndarray_obj_t *out = create_new_ndarray(1, len, NDARRAY_FLOAT);
     mp_float_t *outptr = out->array->items;
     int off = len_c-1;
-    for(int k=-off; k<len-off; k++) {
-        mp_float_t accum = (mp_float_t)0;
-        int top_n = MIN(len_c, len_a - k);
-        int bot_n = MAX(-k, 0);
-        for(int n=bot_n; n<top_n; n++) {
-            int idx_c = len_c - n - 1;
-            int idx_a = n+k;
-            mp_float_t ai = (mp_float_t)0, ci = (mp_float_t)0;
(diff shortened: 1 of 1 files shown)


v923z:
> 
> 
> Special casing floats decreases runtime to about 45% of original (applying a 117-tap filter to 512 points of data goes from 70ms to 32ms)
> 
> The top_n/bot_n calculations already meant that the a/c indices were never out of range. This decreases runtime further to about 15% of original (11ms)
> 
> Timings done on an Adafruit Clue (nrf52840 at 64MHz)
> 
> It does of course increase code size somewhat. It would be possible to put the all-float case inside an `ifdef` if that's a tradeoff you want to offer to users of ulab.

I am not sure. You have just found a factor of 6 speed-up. Previously, the whole `convolve` code cost around 800 bytes. It might be 1.5 kB now, but it is blazing fast, so I would just accept it as it is. 

As for the code size: this is again, one of those cases, where two `ndarrays` have to be processed at the same time, so treating all cases scales quadratically. I would have two comments on that. 

1. I was thinking of yanking the int8, and int16 types, and bringing in complexes. With that we would be down to 4 base types (uint8, uint16, float/double, complex), so we could gain something like 30% in size on all code that involves two arrays. That would solve the issue with `fft`, and we could also remove `spectrum`, because that would simply be equivalent to `abs(fft(...))`. 

2. We could think about working with floats for uint8, uint16, and float, and cast the end result to the upcast type. By doing so, we could simply call `ndarray_get_float_value`, which would remove the quadratic scaling. It would, of course, affect performance, because of the function call. I don't know off-hand, how expensive that is, we would have to measure. If we go down this path, then we don't even have to discard the int8/int16 types.

PR #36, opened 2020-02-12
filter_convolve: fix build error

jepler:
My earlier change introduced a build error, at least under circuitpython.  I'm not sure why I didn't find this before making the pull request, sorry for the trouble.


v923z:
No problem, and thanks for fixing it!

PR #37, opened 2020-02-12
Make portable to CircuitPython within same codebase

jepler:
Testing performed: Built both circuitpython (ulab PR branch, unix port) and micropython (master branch, unix port) with the same ulab sources.  Ran basic smoketest of creating an ndarray via the ulab.array constructor, including dtype keyword argument.


v923z:
Jeff, this looks good to me. I will just have to remind myself to stick to the convention now;)

PR #39, opened 2020-02-16
Perform continuous integration using github actions

jepler:
On Discord, I mentioned the idea of automatically testing ulab using github actions.

This pull request will build micropython's master branch with ulab support, and then run the tests in tests/

At this time, there's only one test and it's not very useful.  However, new tests can be added by creating a ".py" and a ".py.exp" file.  Then, on future changes such as pushes and pull requests, information about the success or failure can be found in the "actions" tab, or attached to the pull request page itself.

When ulab and numpy are "similar", there's a way of testing that doesn't involve ".exp" files but instead  would run numpy under python3.  After adding steps to install numpy, a test would look similar to
```python
try:
 import ulab as np
except:
 import numpy as np

print(np.eye(3).shape)
print(list(np.eye(3).flatten))
```


jepler:
Here's an example of how a successful run looks: https://github.com/jepler/circuitpython-ulab/runs/448399091?check_suite_focus=true

v923z:
@jepler Jeff, thanks a lot! I haven't forgotten about this, I just haven't got the time to get around it. This is really great, and well worth the effort (it is easy for me to say, because I haven't mounted any:).

PR #41, opened 2020-02-19
Split ulab into multiple modules

jepler:
When client code is written like
```python
from ulab import fft
...
fft.fft(...)
```
and a particular target disables fft at compile time, the error will occur early (at 'import' time), not late when the fft functionality is actually used.

There are still some oddities, but they seem to be limitations of how built in C modules work in micro/circuit python.  Before we take the split downstream in circuitpython, we're going to investigate what can be done to fix it.  Two main things are:  We would prefer if `import ulab` behaved more like a "regular" package, and didn't make `ulab.fft` available.  But, relatedly, `import ulab.fft as fft` doesn't work, giving an odd error:
```
>>> import ulab.fft as fft
ImportError: no module named 'ulab'
```


jepler:
Right now this revives the problem of non working properties in micropython.  For the benefit of micropython users, it should probably not be merged in this state.  SInce we are using our own "bindings" code in circuitpython, one possibility is to revert the use of properties in micropython-ulab.  I'm open to suggestions about how to handle it.

jepler:
Other than that, I think this is good to go.

v923z:
> 
> 
> Right now this revives the problem of non working properties in micropython. For the benefit of micropython users, it should probably not be merged in this state. 

In micropython, properties are not working, anyway, so they have to be called as proper methods now. I hope that this will soon be fixed upstream: https://github.com/micropython/micropython/pull/5648. But I can hold off with merging till this is sorted out. 

> SInce we are using our own "bindings" code in circuitpython, one possibility is to revert the use of properties in micropython-ulab. I'm open to suggestions about how to handle it.

No, I wouldn't insist on that. This is a minor inconvenience in micropython-ulab, and we will find a solution.

v923z:
> There are still some oddities, but they seem to be limitations of how built in C modules work in micro/circuit python. Before we take the split downstream in circuitpython, we're going to investigate what can be done to fix it. Two main things are: We would prefer if `import ulab` behaved more like a "regular" package, and didn't make `ulab.fft` available. But, relatedly, `import ulab.fft as fft` doesn't work, giving an odd error:
> 
> ```
> >>> import ulab.fft as fft
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ImportError: no module named 'ulab'
> ```

That's right. One has to import as

```
from ulab import fft
```
But I also think that you are right, and this issue is actually not in our hands. I will try to look into this, though, I am not so hopeful.

jepler:
Micropython has a PR to make this better: https://github.com/micropython/micropython/pull/5701
and so does CircuitPython: https://github.com/adafruit/circuitpython/pull/2657 -- we will merge this assuming that later on `import ulab.fft` will begin to work.

PR #45, opened 2020-02-27
Move zeros(), ones() to base ulab module

jepler:
These functions should be available regardless of whether linear algebra functions like ulab.linalg.det are available.


v923z:
> These functions should be available regardless of whether linear algebra functions like ulab.linalg.det are available.

I agree with the linalg-independent part. However they don't necessarily have to be available at all. `zeros`, `ones`, `eye`, and `linspace` are initialisation functions that serve as shortcuts, nothing more. We could move the implementations into `construct.c`, `initialise.c`, or something similar, and decide later, whether we bind them at the top level in `ulab`, or in a python sub-module.

PR #48, opened 2020-02-27
created new create sub-module for ndarray initialisation functions

v923z:
This modification simply factors out a couple of initialisation functions into a separate file.


v923z:
I could successfully compile it on the local computer, i have no idea, why the checks have failed...

PR #52, opened 2020-03-02
Deleting "#define MODULE_ULAB_ENABLED (1)"

codemee:
Since the micropython.mk file has added the following line

```
CFLAGS_EXTRA = -DMODULE_ULAB_ENABLED=1
```

There's no need to add #define MODULE_ULAB_ENABLED (1) in the mpconfigport.h.Or it would make redefined errors while compiling code.


v923z:
OK, it's clear now. Thanks for the PR!

PR #56, opened 2020-03-07
include only code changes in workflow file

v923z:
This PR modifies the workflow trigger, so that CI runs only, if files in either code/, or tests/ change. With that, updates to the documentation would not consume github resources.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,10 @@ name: Build CI
 
 on:
   push:
-	paths:
-	- 'code/'
-	- 'tests/'
-	- '.github/workflows/'
+    paths:
+    - 'code/'
+    - 'tests/'
+    - '.github/workflows/'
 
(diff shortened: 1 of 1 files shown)


jepler:
Interesting capability I didn't know about.  CircuitPython project doesn't seem to worry much about usage of github resources for their CI, so I didn't consider it here either.

dhalbert:
We would want to check even doc-only PR's in CircuitPython because there might be a sphinx error that would break the doc part of the build.

v923z:
> 
> 
> We would want to check even doc-only PR's in CircuitPython because there might be a sphinx error that would break the doc part of the build.

This is a good point. However, you can still separate the doc build from the firmware build. As far as I see, you run the whole test suite, even if you change only the documentation. You already do that for the website: https://github.com/adafruit/circuitpython/blob/master/.github/workflows/create_website_pr.yml

PR #68, opened 2020-03-30
Boolean

v923z:
fixes https://github.com/v923z/micropython-ulab/issues/67

PR #69, opened 2020-03-30
Add to #68 for CircuitPython compatibility

jepler:
Also adds a new test.  The test passes both on micropython and circuitpython.


v923z:
> Also adds a new test. The test passes both on micropython and circuitpython.

Great, many thanks!

Issue #14, opened 2019-11-12
Wish list


v923z:
This is the place, where suggestions related to new functions/functionality and the like should be brought up.

v923z:
Add properties, like `.T`, `.dtype`, `.shape` and the like to the `ndarray` class.

v923z:
Let `ndarray`s be of arbitrary dimension.

v923z:
Add support for matrix multiplication, `@`, operator as described in https://github.com/v923z/micropython-ulab/issues/11

v923z:
Let `ndarray`s be on the right hand side of binary operators as described in https://github.com/v923z/micropython-ulab/issues/10

mdaeron:
Allow indexing an array with another array of integers or booleans, similar to `numpy`:

```
from numpy import array
a = array([1, 2, 3])

k = array([2, 0, 1, 1, 0, 2, 2, 1, 0])
print(a[k]) # yields array([3, 1, 2, 2, 1, 3, 3, 2, 1])

k = array([True, False, True])
print(a[k]) # yields array([1, 3])
```

SummerGGift:
I find ur micropython-usermod series articles today, ur works is great, it's very helpful. 

I have used micropython in my work for about two years, during the time, couldn't find many useful articles describe the details in it. i think i will translate some chapters into chinese, and spread them in China. 

Thank you very much.

v923z:
> I find ur micropython-usermod series articles today, ur works is great, it's very helpful.
> 
> I have used micropython in my work for about two years, during the time, couldn't find many useful articles describe the details in it. i think i will translate some chapters into chinese, and spread them in China.
> 

Thank you for your feedback! Since I put the material deliberately into the public domain, I have absolutely nothing against your translating the text.

roger-:
How about a circular buffer compatible with ndarray (but fixed size of course)? Would avoid all the memmoves needed with roll().

v923z:
This is definitely true, in fact, I also wanted to implement this. This would be extra compared to numpy, right? Actually, this is why I chose the name `ulab` and not `unumpy` or something similar.

Basically, you want a sub-class of ndarray, with a special method, say, `insert`, which would move the pointer. The question is, perhaps, what would you do with the numerical methods? Some do not care about the order of the elements (sum, mean, std, FFT, etc.), while some do (argmin, polyfit, polyval). One other problem is slicing. If you have any suggestions, I am listening.

roger-:
[...]

PR #75, opened 2020-04-01
adds an implementation of the `**` binary operator, and the reversed binary operators

v923z:
The implementation of `**` should answer the feature request in https://github.com/v923z/micropython-ulab/issues/14#issuecomment-605499572. Reversed operators (i.e., when the `ndarray` stands on the right hand side of an operator, as in 
```
a = array([1, 2, 3])
a + 1
1 + a
```
were requested a long time ago. 

A couple of examples: 
```
import ulab
a = ulab.array(range(10), dtype=ulab.int8)
print(1+a, a+1)
print(2*a, a*2)
print(2**a, a**2)
print(a-1, 1-a)
print(a/2, 2/a)
```


v923z:
> 
> 
> Looks good to me, but github demands tribute (resolution of merge conflicts) before it can be merged.

Thanks! It turns out that only the version number and the change log had conflicts.

PR #86, opened 2020-04-06
Restore circuitpython compatibility after #84

jepler:
#84 makes use of a recently added micropython feature.  Use preprocessor tests to restore compatibility when building in CircuitPython.

### DIFF
--- a/code/ulab.c
+++ b/code/ulab.c
@@ -50,7 +50,9 @@ STATIC MP_DEFINE_CONST_DICT(ulab_ndarray_locals_dict, ulab_ndarray_locals_dict_t
 
 const mp_obj_type_t ulab_ndarray_type = {
     { &mp_type_type },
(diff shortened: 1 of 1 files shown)


jepler:
Note that this means ulab-in-CircuitPython and ulab-in-micropython currently have different behaviors, until CircuitPython implements the new flags:

Circuitpython:
```
>>> ones(3) == 1
False
```

Micropython:
```
>>> ones(3) == 1
[True, True, True]
```
However, since the detection is by #if defined(), when circuitpython does this no action is needed in ulab.

dhalbert:
We should see about cherry-picking the MicroPython change into CircuitPython.

v923z:
@jepler this is absolutely OK, you could've merged it.

v923z:
A workaround for `circuitpython` is provided in https://github.com/v923z/micropython-ulab/pull/115

PR #88, opened 2020-04-06
Fix compiler diagnostics normally enabled in CircuitPython; fix arg parsing bugs

jepler:
CircuitPython enables a large number of compiler diagnostics.  Initially, we disabled these diagonstics for ulab only.  However, the vast majority of them were very feasible to fix, and a few bugs in argument handling were discovered along the way.

This includes #86 since it allowed me to test both MicroPython and CircuitPython with my changes.

The following diagnostic flags can be reenabled in CircuitPython after this is included:

> -Werror=sign-compare -Werror=missing-prototypes -Werror=unused-parameter -Werror=missing-declarations -Werror=shadow -Werror=maybe-uninitialized

This fixes bugs including that `linspace(0, 1, 9)` generated an array of length 50, since the third positional argument was not sent to the "num" parameter.


code/ndarray.c
```diff
@@ -281,13 +282,13 @@ size_t true_length(mp_obj_t bool_list) {
...
 
-mp_bound_slice_t generate_slice(mp_uint_t n, mp_obj_t index) {
+static mp_bound_slice_t generate_slice(mp_uint_t n, mp_obj_t index) {
     // micropython seems to have difficulties with negative steps
     mp_bound_slice_t slice;
     if(MP_OBJ_IS_TYPE(index, &mp_type_slice)) {
         mp_obj_slice_indices(index, n, &slice);
     } else if(MP_OBJ_IS_INT(index)) {
-        int32_t _index = mp_obj_get_int(index);
+        size_t _index = mp_obj_get_int(index);
```
v923z, line 291: I think I should have commented here. int32_t is probably necessary, since _index < 0 is possible.

v923z:
@jepler many thanks for weeding out a number of errors! My only comment is on the nature of `_index`.

code/ndarray.c
jepler: Fixed in 86e7155ab7b72ce3cc63d29b4b053f5095061412

v923z:
Thanks, Jeff!

jepler:
Thank you for looking at it with a second pair of eyes

Issue #87, opened 2020-04-06
argmin/argmax of iterable: results differ from ndarray and numpy


jepler:
```
>>> ulab.numerical.argmax([[1,2],[4,3]])
1
>>> ulab.numerical.argmin([[1,2],[4,3]])
0
```
but
```
>>> ulab.numerical.argmin(ulab.array([[1,2],[4,3]]))
0
>>> ulab.numerical.argmax(ulab.array([[1,2],[4,3]]))
2
```
and
```
>>> numpy.argmin([[1,2],[4,3]])
0
>>> numpy.argmax([[1,2],[4,3]])
2
```

v923z:
This is sort of OK, and intentional: argmax and similar functions will work with one-dimensional iterables only: https://github.com/v923z/micropython-ulab/blob/280f04660e4ee7b56117c21bfe87a0274429e965/code/numerical.c#L151 
We should probably throw an exception, if this is not the case. The same applies to the other numerical functions: https://github.com/v923z/micropython-ulab/blob/280f04660e4ee7b56117c21bfe87a0274429e965/code/numerical.c#L60

If one really needs the maximum of a list of lists, then it should be converted to an `ndarray` first, thereby delegating the inspection to the initialisation function. Do you think that lists of lists should be supported here?

jepler:
No, I don't have a strong opinion. I looked at this code because there was an unused argument diagnostic nearby. The behavior isn't actually anyone to the best of my knowledge.

v923z:
> No, I don't have a strong opinion. I looked at this code because there was an unused argument diagnostic nearby. The behavior isn't actually anyone to the best of my knowledge.

I think at least 1D iterables should be supported, and the question really is how we prevent accidents like your example from happening. An easy way out would be to call `mp_obj_get_float`, and then compare numbers, instead of using `mp_binary_op(...)`.

And I have to go back on my earlier comment: `std`, `mean` and the like will not produce the error you found, because they do call `mp_obj_get_float`, so an exception will be raised automatically, if you try to pass lists of lists to the function.

v923z:
@jepler https://github.com/v923z/micropython-ulab/pull/91 is an attempt at a solution. If you agree, you can merge it, otherwise let me know if you have other suggestions.

PR #91, opened 2020-04-07
fixed argmin/argmax function for iterables

v923z:
This PR should fix https://github.com/v923z/micropython-ulab/issues/87: if items of an iterable cannot be converted to floats, a `TypeError` exception will be raised automatically.


v923z:
> I think it also needs handling of empty sequences.
> 
> Before this change, the answer to `argmin([])` was nonsense, but now it causes a segmentation fault.

Thanks for pointing this out, I will see to it.
Unfortunately, `numpy` itself is a bit inconsistent as far as arg-anything is concerned: `argmin` raises a `ValueError` with an empty sequence, while `argsort` returns an empty array. Corner cases are always a bit annoying.

v923z:
> I think it also needs handling of empty sequences.
> 
> Before this change, the answer to `argmin([])` was nonsense, but now it causes a segmentation fault.

`numpy` raises separate `ValueError`s for `argmin` and `argmax`, respectively. I treated them with a single exception, namely, "attempt to get argmin/argmax of an empty sequence". If you insist, we can definitely change that, but I don't think it is worth the extra 100 bytes that it would cost.

jepler:
I wrote a test that still shows some weirdness with argmin/max
```
import ulab
    
# Adapted from https://docs.python.org/3.8/library/itertools.html#itertools.permutations
def permutations(iterable, r=None):
    # permutations('ABCD', 2) --> AB AC AD BA BC BD CA CB CD DA DB DC
    # permutations(range(3)) --> 012 021 102 120 201 210
    pool = tuple(iterable) 
    n = len(pool)
    r = n if r is None else r
    if r > n: 
        return
    indices = list(range(n))
    cycles = list(range(n, n-r, -1))
    yield tuple(pool[i] for i in indices[:r])
    while n:
        for i in reversed(range(r)):
            cycles[i] -= 1
            if cycles[i] == 0:
                indices[i:] = indices[i+1:] + indices[i:i+1]
                cycles[i] = n - i
            else:
                j = cycles[i]
                indices[i], indices[-j] = indices[-j], indices[i]
                yield tuple(pool[i] for i in indices[:r])
                break
        else:
            return

# Combinations expected to throw
try:
    print(ulab.numerical.argmin([]))
except ValueError:
    print("ValueError")

try:
    print(ulab.numerical.argmax([]))
except ValueError:
    print("ValueError")

# Combinations expected to succeed
print(ulab.numerical.argmin([1]))
print(ulab.numerical.argmax([1]))
print(ulab.numerical.argmin(ulab.array([1])))
print(ulab.numerical.argmax(ulab.array([1])))

print()
print("max tests")
for p in permutations((100,200,300)):
    m1 = ulab.numerical.argmax(p)
    m2 = ulab.numerical.argmax(ulab.array(p))
    print(p, m1, m2)
    if m1 != m2 or p[m1] != max(p):
        print("FAIL", p, m1, m2, max(p))
    
print()
print("min tests")
for p in permutations((100,200,300)):
    m1 = ulab.numerical.argmin(p)
    m2 = ulab.numerical.argmin(ulab.array(p))
    print(p, m1, m2)
    if m1 != m2 or p[m1] != max(p):
        print("FAIL", p, m1, m2, max(p))
```

v923z:
Thanks for the test!  I have implemented a fix. I hope this is OK now.

> print("min tests")
> for p in permutations((100,200,300)):
>     m1 = ulab.numerical.argmin(p)
>     m2 = ulab.numerical.argmin(ulab.array(p))
>     print(p, m1, m2)
>     if m1 != m2 or p[m1] != max(p):
>         print("FAIL", p, m1, m2, max(p))
> ```

I think the last two lines should read as 

```
if m1 != m2 or p[m1] != min(p):
   print("FAIL", p, m1, m2, min(p))
```

PR #101, opened 2020-04-20
polyfit: Argument handling improvements

jepler:
* In the 3-args case, the lengths of the arguments were not checked
 * in the 3-args case, the type of the 2nd argument was not checked
 * gcc falsely diagnosed a `maybe-uninitialized` variable because it did not see that the branches of the if() statement were mutually exclusive

It's this third issue that originally drew my attention to this code, adafruit/circuitpython#2787


v923z:
>     * In the 3-args case, the lengths of the arguments were not checked
> 
>     * in the 3-args case, the type of the 2nd argument was not checked

Uh, this is a bad error, thanks for catching it! I will just go ahead, and merge your modifications.

PR #108, opened 2020-04-25
Added support for BLAS saxpy and sdot

JishinMaster:
I converted some BLAS functions with f2c, saxpy and sdot, and added them to the extra module.
If MICROPY_FLOAT_IMPL is MICROPY_FLOAT_IMPL_DOUBLE it computes daxpy and ddot.

I tried it out on the micropython unix port.
It would be interesting to also have sgemv, sgemm, strsm, and their complex counterpart.

Please let me know if you need code refactoring.

### DIFF
--- a/code/blas.h
+++ b/code/blas.h
@@ -1,0 +1,163 @@
+#pragma once
+#include "f2c.h"
+
+
+#if(MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE)
+#define ZERO 0.0
+#else
+#define ZERO 0.0f
+#endif
+
+/* Subroutine */ int axpy_(const integer *n, const mp_float_t *da, const mp_float_t *dx, 
+	const integer *incx, mp_float_t *dy, const integer *incy)
+{
+    /* System generated locals */
+    integer i__1;
+
+    /* Local variables */
+    static integer i__, m, ix, iy, mp1;
+
+    --dy;
+    --dx;
+
+    /* Function Body */
+    if (*n <= 0) {
+	return 0;
+    }
+    if (*da == ZERO) {
+	return 0;
+    }
+    if (*incx == 1 && *incy == 1) {
+	m = *n % 4;
+	if (m != 0) {
+	    i__1 = m;
+	    for (i__ = 1; i__ <= i__1; ++i__) {
+		dy[i__] += *da * dx[i__];
+	    }
+	}
+	if (*n < 4) {
+	    return 0;
+	}
+	mp1 = m + 1;
+	i__1 = *n;
+	for (i__ = mp1; i__ <= i__1; i__ += 4) {
+	    dy[i__] += *da * dx[i__];
+	    dy[i__ + 1] += *da * dx[i__ + 1];
+	    dy[i__ + 2] += *da * dx[i__ + 2];
+	    dy[i__ + 3] += *da * dx[i__ + 3];
+	}
+    } else {
+
+/*        code for unequal increments or equal increments */
+/*          not equal to 1 */
+
+	ix = 1;
+	iy = 1;
+	if (*incx < 0) {
+	    ix = (-(*n) + 1) * *incx + 1;
+	}
+	if (*incy < 0) {
+	    iy = (-(*n) + 1) * *incy + 1;
+	}
+	i__1 = *n;
+	for (i__ = 1; i__ <= i__1; ++i__) {
+	    dy[iy] += *da * dx[ix];
+	    ix += *incx;
+	    iy += *incy;
+	}
+    }
+    return 0;
+} /* axpy_ */
+
+void cblas_axpy( const int N, const mp_float_t alpha, const mp_float_t *X,
+                       const int incX, mp_float_t *Y, const int incY)
+{
+   long int F77_N=N, F77_incX=incX, F77_incY=incY;
+   axpy_( &F77_N, &alpha, X, &F77_incX, Y, &F77_incY);
+}
+
+///SDOT
+mp_float_t sdot_(const integer *n, const mp_float_t *sx, const integer *incx, const mp_float_t *sy, const integer *incy)
+{
+    /* System generated locals */
+    integer i__1;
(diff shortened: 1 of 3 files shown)


v923z:
Thanks for the PR!

A couple of comments:

1. The dot product is already supported in `ulab`: https://github.com/v923z/micropython-ulab/blob/314bb010bb0282888b21c20123eba5d87f20874f/code/linalg.c#L142 Does `sdot` bring in any extra features?

2. The functionality of `saxpy` is already defined, though, I admit, it requires two calls:
```
a = array([1, 2, 3, 4])
b = array([5, 6, 7, 8])
c = a*12 + b
```
You can find the implementation in https://github.com/v923z/micropython-ulab/blob/314bb010bb0282888b21c20123eba5d87f20874f/code/ndarray.c#L739. Given that microcontrollers don't have unlimited resources, I wouldn't like to include stuff that only duplicates implementation.

3. And this is, perhaps, my strongest reason: both implementations are restrictive, and can't deal with integer types. The available implementations in `ulab` do that. I believe, all functions should be able to deal with arbitrary `ndarray`s, unless there is a very pressing reason for not doing so.

4. I don't know, what the license terms of BLAS are, but we should not include anything that is stricter than the MIT license.

5. `ulab` is included in many projects now, partly because it has absolutely no dependencies. I believe, it should remain so. Actually, given the simplicity of the `sdot` and `saxpy` routines, I don't even think that it would be necessary to include BLAS: everything can be implemented in vanilla C. If `saxpy` should be included, that is.

JishinMaster:
Thank you for your quick reply!

I am accustomed to BLAS and LAPACK functions, which I use in C/C++ code.
I know that Numpy (and Matlab) use the BLAS functions transparently, but I would have like an BLAS/LAPACK like API.

The Netlib implementation is supposedly faster than a naive C one (and it takes into account all the special cases). OpenBLAS/ATLAS/etc are way faster but use architecture optimisations, such as SIMD C intrinsics or assembly.

The Netlib BLAS license is located here, and has : http://www.netlib.org/blas/#_licensing
_"The reference BLAS is a freely-available software package. It is available from netlib via anonymous ftp and the World Wide Web. Thus, it can be included in commercial software packages (and has been). We only ask that proper credit be given to the authors.
Like all software, it is copyrighted. It is not trademarked, but we do ask the following:
 -  If you modify the source for these routines we ask that you change the name of the routine and comment the changes made to the original.
-  We will gladly answer any questions regarding the software. If a modification is done, however, it is the responsibility of the person who modified the routine to provide support.
"_

I had issues building micropython on unix with mp_float_t as float. It seems to be related to make options, which I did not want to change. The differences between the sdot/ddot and saxpy/daxpy are the input/output type (mp_float_t in both cases, which is either float or double depending on the micropython configuration file), and float initialisation variable, which are either 0.0f (float) or 0.0.

It seems indeed that your linalg_dot function does a more complete work than the BLAS-Level1 sdot, since it can work with more than float, and more than 1D arrays. The complex functions, cdot/zdot, do not have yet an equivalent.

Thank you for clarifying the standalone aspect of ulab, I had the opposite idea, micropython wrapping small optimised C only libraries (cblas/blas for linear algeabra, BearSSL for crypto, fftw for FFT/DFT, etc).

Do you think it could be interesting as a separate micropython experimental library?

Should I close the pull request?

v923z:
Many thanks for the feedback!

> The Netlib implementation is supposedly faster than a naive C one (and it takes into account all the special cases). OpenBLAS/ATLAS/etc are way faster but use architecture optimisations, such as SIMD C intrinsics or assembly.

If we can gain in speed without sacrificing too much RAM, we could think about bringing in BLAS. However, it must be done in a way that does not break compatibility with already supported hardware. 

I also considered assembly, but I discarded it for two reasons. One is that I don't know that well enough, and second, that would definitely have raised the bar for anyone wanting to contribute. 

And on a microcontroller, the constraints are a bit different to those on a PC. E.g., the FFT could be done slightly faster than it is done in `ulab`, but you would need extra RAM for that. The question one has to answer is, whether 20% gain in speed is worth doubling the RAM requirements. I had the feeling that I would have more control over such things, if no external library was involved, even if it meant that we have to develop our own routines. Which was actually quite instructive.
> 
> I had issues building micropython on unix with mp_float_t as float. It seems to be related to make options, which I did not want to change. The differences between the sdot/ddot and saxpy/daxpy are the input/output type (mp_float_t in both cases, which is either float or double depending on the micropython configuration file), and float initialisation variable, which are either 0.0f (float) or 0.0.

On the unix port, `mp_float_t` is double.

> 
> It seems indeed that your linalg_dot function does a more complete work than the BLAS-Level1 sdot, since it can work with more than float, and more than 1D arrays. The complex functions, cdot/zdot, do not have yet an equivalent.

The reason for that is that `ulab` does not have a complex type. I don't know, whether it is really necessary, except for FFT, and, on the other hand, it adds a lot to the complexity and size of the firmware. (E.g., the size of the binary operator code, or anything, where two arrays are concerned, scales with the square of the number of supported types. It makes a big difference, whether you have 5, or 6.) 

> 
> Thank you for clarifying the standalone aspect of ulab, I had the opposite idea, micropython wrapping small optimised C only libraries (cblas/blas for linear algeabra, BearSSL for crypto, fftw for FFT/DFT, etc).
> 
> Do you think it could be interesting as a separate micropython experimental library?

That is certainly a different approach, and I don't know the answer. I would, however, point out that it is actually not the algorithm implementations that are difficult in `numpy`, but the parsing, and bringing the inputs into a form that you can feed into the standard algorithms. You can't possibly bypass that, no matter which library you use. 

And you would still need an efficient container for the data that you produce. `numpy`'s `ndarray` seemed standard in `python`, so that is what I chose.

One last comment: BLAS, and `numpy`, or any other math libraries that you can find on computer, can much more than what you would possibly need on a microcontroller, so you would probably throw 99% of the library away.

> Should I close the pull request?

It's up to you. However, if you think that you would need a function from BLAS within the framework of `ulab`, I am open to suggestions.

PR #110, opened 2020-04-27
ulab.h: Actually allow excluding modules

jepler:
Perhaps I'm overlooking something, but when I tried to _actually_ disable a module by adding
`#define ULAB_COMPARE_MODULE (0)` in an mpconfigboard.h, I got multiple diagnostics like
```
../../extmod/ulab/code/ulab.h:37: error: "ULAB_COMPARE_MODULE" redefined [-Werror]
```

Therefore, in ulab.h, check whether the enabling-symbol is defined or not before defining it to (1) to enable everything by default.

This is not done for ULAB_CREATE_MODULE since, as the comments note, it is always enabled.

We may need to use this in CircuitPython soon because, at the moment, adding ulab.compare overflows one build, the espruino pico (stm).  See adafruit/circuitpython#2811 for progress integrating the current (or now slightly old :frowning:) ulab.

### DIFF
--- a/code/ulab.h
+++ b/code/ulab.h
@@ -16,27 +16,43 @@
 #define ULAB_CREATE_MODULE (1)
 
 // vectorise (all functions) takes approx. 4.5 kB of flash space
+#ifndef ULAB_VECTORISE_MODULE
 #define ULAB_VECTORISE_MODULE (1)
+#endif
 
 // linalg adds around 6 kB to the firmware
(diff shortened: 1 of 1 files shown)


v923z:
Thanks, Jeff! On `micropython`, I didn't see this problem (I actually measure the cost of a module by simply setting the flag to 0), but if it is required on `circuitpython` then `ulab.h` must obviously be done as you suggest.

PR #121, opened 2020-06-01
fixes min/max error from adafruit/circuitpython #2984

v923z:
This PR should fix the issue mentioned in https://github.com/adafruit/circuitpython/issues/2984


v923z:
> makes sense, didn't test

I have run a couple of tests by hand.

PR #122, opened 2020-06-01
Updates needed to build latest code on CircuitPython

jepler:
This fixes the following build error when updating ulab in CircuitPython:
```
../../extmod/ulab/code/vectorise.c: In function 'vectorise_vectorized_function_call':
../../extmod/ulab/code/vectorise.c:242:15: error: implicit declaration of function 'mp_obj_is_int'; did you mean 'mp_obj_list_init'? [-Werror=implicit-function-declaration]
  242 |     } else if(mp_obj_is_int(args[0]) || mp_obj_is_float(args[0])) {
      |               ^~~~~~~~~~~~~
      |               mp_obj_list_init
../../extmod/ulab/code/vectorise.c:242:15: error: nested extern declaration of 'mp_obj_is_int' [-Werror=nested-externs]
cc1: all warnings being treated as errors
```

### DIFF
--- a/code/ndarray.h
+++ b/code/ndarray.h
@@ -28,6 +28,7 @@
 
 #if CIRCUITPY
 #define mp_obj_is_bool(o) (MP_OBJ_IS_TYPE((o), &mp_type_bool))
+#define mp_obj_is_int(x) (MP_OBJ_IS_INT((x)))
 #else
 #define translate(x) MP_ERROR_TEXT(x)
 #endif
--- a/code/approx.c
+++ b/code/approx.c
@@ -22,7 +22,7 @@
 const mp_obj_float_t xtolerance = {{&mp_type_float}, 2.4e-7};
 const mp_obj_float_t rtolerance = {{&mp_type_float}, 0.0};
 
-mp_float_t approx_python_call(const mp_obj_type_t *type, mp_obj_t fun, mp_float_t x, mp_obj_t *fargs, uint8_t nparams) {
+STATIC mp_float_t approx_python_call(const mp_obj_type_t *type, mp_obj_t fun, mp_float_t x, mp_obj_t *fargs, uint8_t nparams) {
 	// Helper function for calculating the value of f(x, a, b, c, ...), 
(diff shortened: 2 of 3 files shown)


v923z:
Won't we run into circular definitions? This is from obj.h: 
https://github.com/micropython/micropython/blob/22806ed5df27c10131af0cedb2f7e8b134fe6e7a/py/obj.h#L1005

I have to admit, I am never sure, whether one should use lower case, or upper case with these type checks. If you know what to do, I am not against making the whole thing uniform, so that we wouldn't have to distinguish between `circuitpython`, and `micropython`.

One might have to consider, though, that macros are in-lined, so we would probably add to the size of the firmware

jepler:
If CircuitPython ever updates from MicroPython, we'll have to change it.  But for now, this seems like the easiest way.

I've updated this PR with additional fixes for building on CircuitPython, due to our compiler flags.

v923z:
> If CircuitPython ever updates from MicroPython, we'll have to change it. But for now, this seems like the easiest way.
> 
> I've updated this PR with additional fixes for building on CircuitPython, due to our compiler flags.

OK. Go ahead!

code/ndarray.h
```diff
@@ -28,6 +28,7 @@
 
 #if CIRCUITPY
 #define mp_obj_is_bool(o) (MP_OBJ_IS_TYPE((o), &mp_type_bool))
+#define mp_obj_is_int(x) (MP_OBJ_IS_INT((x)))
```
dhalbert, line 31: I agree, this is the easiest way. MicroPython was just trying to canonicalize their macro names.

dhalbert: I didn't mean for this to be a review, just a comment.

v923z: Oh, but a comment by the oracle qualifies as a review. Seriously, Dan, I really appreciate that you chipped in.

dhalbert: I am only an oracle because I subscribe to all the micropython git notifcations :smile:

PR #123, opened 2020-06-01
github actions: Fix paths directives

jepler:
I noticed on #122 that no actions were run.  I believe this is because the paths directives had no wildcards, contrary to the examples at https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

Eliminating the paths: restrictions entirely may be an equally valid course of action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,9 +4,9 @@ on:
   push:
   pull_request:
     paths:
-    - 'code/'
-    - 'tests/'
-    - '.github/workflows/'
+    - 'code/**'
+    - 'tests/**'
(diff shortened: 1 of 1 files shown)


jepler:
I may be confused about what is going on here.  #121 had its actions run, #122 did not.

v923z:
> I may be confused about what is going on here. #121 had its actions run, #122 did not.

I think you are right, `**` is required: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions

v923z:
@jepler Jeff, thanks for getting to the bottom of this!

PR #141, opened 2020-07-21
fix arange : missing MICROPY_FLOAT_C_FUN(ceil) in create.c

rcolistete:
Fix for 'ulab.arange()' code in 'create.c', due to missing MICROPY_FLOAT_C_FUN(ceil).

So it compiles not only on 'ports/unix' and on 'ports/stm32' (etc) with DP (double precision), but also on 'ports/stm32' (etc) with SP (single precision).


v923z:
Thanks, this is indeed an error.

PR #143, opened 2020-07-22
linalg.dot : also accepts 2 vectors of same dimension

rcolistete:
Extend 'ulab.linalg.dot' to accept 2 vectors, so it becomres more similar to [`numpy.dot`](https://numpy.org/doc/stable/reference/generated/numpy.dot.html?highlight=numpy%20dot) and the 1st case from the documentation will be accepted :
- If both a and b are 1-D arrays, it is inner product of vectors (without complex conjugation).
besides the 2nd one :
If both a and b are 2-D arrays, it is matrix multiplication, ...

### DIFF
--- a/code/linalg.c
+++ b/code/linalg.c
@@ -145,26 +145,40 @@ static mp_obj_t linalg_dot(mp_obj_t _m1, mp_obj_t _m2) {
         mp_raise_TypeError(translate("arguments must be ndarrays"));
     }
     ndarray_obj_t *m1 = MP_OBJ_TO_PTR(_m1);
-    ndarray_obj_t *m2 = MP_OBJ_TO_PTR(_m2);    
-    if(m1->n != m2->m) {
-        mp_raise_ValueError(translate("matrix dimensions do not match"));
-    }
-    // TODO: numpy uses upcasting here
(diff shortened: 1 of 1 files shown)


code/linalg.c
```diff
@@ -145,26 +145,40 @@ static mp_obj_t linalg_dot(mp_obj_t _m1, mp_obj_t _m2) {
...
-                // (i, k) * (k, j)
-                v1 = ndarray_get_float_value(m1->array->items, m1->array->typecode, i*m1->n+k);
-                v2 = ndarray_get_float_value(m2->array->items, m2->array->typecode, k*m2->n+j);
-                sum += v1 * v2;
+    ndarray_obj_t *m2 = MP_OBJ_TO_PTR(_m2); 
+
+    if ((m1->m == 1) && (m2->m == 1)) {
+        // 2 vectors
+        if (m1->array->len != m2->array->len) {
+            mp_raise_ValueError(translate("vectors must have same dimension"));
```
v923z, line 153: Could you, please, change the error message to "vectors must have same lengths"? Dimension does not make sense here.

rcolistete: Ok, made a 2nd commit, "linalg.dot : changed error message: vectors must have same lengths".

v923z:
This is OK now, thanks for the contribution!

PR #145, opened 2020-07-22
added linalg.norm : for vector or matrix

rcolistete:
Implemented `ulab.linalg.norm(x)`, where x can be a vector or matrix, without options. Based on [`numpy.linalg.norm`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html).


code/linalg.c
```diff
@@ -183,6 +183,21 @@ static mp_obj_t linalg_dot(mp_obj_t _m1, mp_obj_t _m2) {
...
 
+static mp_obj_t linalg_norm(mp_obj_t _x) {
+    if (!MP_OBJ_IS_TYPE(_x, &ulab_ndarray_type)) {
+        mp_raise_TypeError(translate("argument must be ndarray"));
+    }
+    ndarray_obj_t *x = MP_OBJ_TO_PTR(_x);
+    mp_float_t dot = 0.0, v;
+    for (size_t pos=0; pos < x->array->len; pos++) {
+        v = ndarray_get_float_value(x->array->items, x->array->typecode, pos);
+        dot += v*v;
```
v923z, line 194: Check out, whether `ord` with numerical values makes sense.

code/linalg.c
```diff
@@ -183,6 +183,21 @@ static mp_obj_t linalg_dot(mp_obj_t _m1, mp_obj_t _m2) {
 
 MP_DEFINE_CONST_FUN_OBJ_2(linalg_dot_obj, linalg_dot);
 
+static mp_obj_t linalg_norm(mp_obj_t _x) {
+    if (!MP_OBJ_IS_TYPE(_x, &ulab_ndarray_type)) {
```
v923z, line 187: If it is not very expensive, I would allow generic iterables. That would mean that the `for` loop has to be implemented once more. The rest is OK.

code/linalg.c
rcolistete: What do to with numerical 'ord' ? From [numpy.linalg.norm](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html) :
- for vectors, ord != 0 yields `sum(abs(x)**ord)**(1./ord)`;
- for matrices, it is a mess (in NumPy), with many cases...

v923z: That's right. Let's leave it, then.

v923z:
@rcolistete Roberto, do you think you can resolve the conflicts in the next hour or so? We would like to merge these changes as soon as possible.

rcolistete:
> @rcolistete Roberto, do you think you can resolve the conflicts in the next hour or so? We would like to merge these changes as soon as possible.

To solve the conflict I should move the linalg.* files to "/code/linalg/", right ? As you have added folders for the modules in v0.53.2.

v923z:
> To solve the conflict I should move the linalg.* files to "/code/linalg/", right ? As you have added folders for the modules in v0.53.2.

Right. We re-organised the code a bit, because generating documentation is easier that way.

v923z:
> 
> To solve the conflict I should move the linalg.* files to "/code/linalg/", right ? As you have added folders for the modules in v0.53.2.

You also need to change the #includes, and add your name in the header, and change the version number go 0.54.

rcolistete:
> > To solve the conflict I should move the linalg.* files to "/code/linalg/", right ? As you have added folders for the modules in v0.53.2.
> 
> You also need to change the #includes, and add your name in the header, and change the version number to 0.54.

Doing the changes, wait some few minutes.

v923z:
> Doing the changes, wait some few minutes.

We can add the iterable part later.

code/linalg/linalg.c
```diff
@@ -233,6 +234,29 @@ static mp_obj_t linalg_dot(mp_obj_t _m1, mp_obj_t _m2) {
...
+    mp_float_t dot = 0.0, v;
+    for (size_t pos=0; pos < x->array->len; pos++) {
+        v = ndarray_get_float_value(x->array->items, x->array->typecode, pos);
+        dot += v*v;
+    }
+    return mp_obj_new_float(MICROPY_FLOAT_C_FUN(sqrt)(dot));
+}
+
+MP_DEFINE_CONST_FUN_OBJ_1(linalg_norm_obj, linalg_norm);
+
```
v923z, line 259: Could you, please, place the function after `inv`, so that the documentation is alphabetically ordered?

code/linalg/linalg.c
```diff
@@ -471,6 +495,7 @@ STATIC const mp_rom_map_elem_t ulab_linalg_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR_cholesky), (mp_obj_t)&linalg_cholesky_obj },
 	{ MP_ROM_QSTR(MP_QSTR_det), (mp_obj_t)&linalg_det_obj },
 	{ MP_ROM_QSTR(MP_QSTR_dot), (mp_obj_t)&linalg_dot_obj },
+	{ MP_ROM_QSTR(MP_QSTR_norm), (mp_obj_t)&linalg_norm_obj },
```
v923z, line 498: Here, too.

code/linalg/linalg.h
```diff
@@ -34,6 +34,7 @@ extern mp_obj_module_t ulab_linalg_module;
 MP_DECLARE_CONST_FUN_OBJ_KW(linalg_size_obj);
 MP_DECLARE_CONST_FUN_OBJ_1(linalg_inv_obj);
 MP_DECLARE_CONST_FUN_OBJ_2(linalg_dot_obj);
+MP_DECLARE_CONST_FUN_OBJ_2(linalg_norm_obj);
```
v923z, line 37: And finally, here.

code/linalg/linalg.c
```diff
@@ -233,6 +234,29 @@ static mp_obj_t linalg_dot(mp_obj_t _m1, mp_obj_t _m2) {
...
 MP_DEFINE_CONST_FUN_OBJ_2(linalg_dot_obj, linalg_dot);
 
+//| def norm(x):
+//|    """
+//|    :param ~ulab.array x: a vector or matrix
+//|
+//|    Computes the norm of a vector or matrix."""
+//|    ...
+//|
+
```
v923z, line 244: Here you could be a bit more specific:

//| def norm(x: ulab.array) -> float:
//|    """
//|    :param ulab.array x
//|
//|	The function takes no parameters, and
//|	returns the 2-norm of a matrix or vector, i.e., sqrt(sum(x*x)), however, without the RAM overhead.
//|	"""
//|

rcolistete:
> > Doing the changes, wait some few minutes.
> 
> We can add the iterable part later.

Yes, I'm testing some benchmarks.

rcolistete:
Is it ok ?

code/linalg/linalg.c
rcolistete: Ok.

code/linalg/linalg.c
rcolistete: Ok.

code/linalg/linalg.h
rcolistete: In 'linalg.h' the functions aren't ordered at all. 'norm' is already after 'dot.
What to do ?

v923z: That's right, but the header file doesn't really matter. The documentation will be generated from the //| comments, which are in the .c file. So it is enough, if you move the function and its documentation in linalg.c.

code/linalg/linalg.c
rcolistete: 'no parameters' ? Or options ?

v923z: There aren't any, are there? You call `norm(x)`, without `ord`, or `axis`, or anything else.

rcolistete: It is the notation, for me x is a parameter, ord, axis, etc, would be options.

v923z: x would be the argument, I think, but either way is fine with me.

code/linalg/linalg.c
```diff
@@ -399,6 +400,31 @@ static mp_obj_t linalg_inv(mp_obj_t o_in) {
...
 
+//| def norm(x: ulab.array) -> float:
+//|    """
+//|    :param ~ulab.array x: a vector or a matrix
+//|
+//|    Computes the 2-norm of a vector or a matrix, i.e., sqrt(sum(x*x)), however, without the RAM overhead.."""
+//|    ...
+//|
+
+the 2-norm of a matrix or vector, i.e., sqrt(sum(x*x)), however, without the RAM overhead.
```
v923z, line 411: You have to take this line out, otherwise, the compiler fails.

code/linalg/linalg.c
rcolistete: Commited.

code/linalg/linalg.c
rcolistete: Sorry, paste which should be removed...

Commited.

v923z:
Thanks for jumping in at such a short notice, and fixing the outstanding issues!

PR #154, opened 2020-07-24
Add MICROPY_FLOAT_CONST to all fp constants.

jonathanhogg:
In order to get ulab to compile correctly against single floating point all of the constants need to switch to single format. Conveniently MicroPython has provided a macro to manage this switch. Use this throughout.

Unfortunately also introduces a bunch of whitespace changes because there is a mass of trailing whitespace in the codebase and my editor is (correctly) configured to remove this.


v923z:
> In order to get ulab to compile correctly against single floating point all of the constants need to switch to single format. Conveniently MicroPython has provided a macro to manage this switch. Use this throughout.

Many thanks for bringing the issue up, and especially, for fixing it!

> 
> Unfortunately also introduces a bunch of whitespace changes because there is a mass of trailing whitespace in the codebase and my editor is (correctly) configured to remove this.

That's fine, don't worry about it.

PR #163, opened 2020-08-01
degrees, radians, pi, e

v923z:
This PR adds two functions to vectorise.c and two module constants.


v923z:
All features have been implemented in the tensor branch.

PR #167, opened 2020-08-06
Minor fixes on Python stubs

ciscorn:
I found some paragraphs are mistakenly joined (by lacking empty lines).


v923z:
Many thanks! I haven't had time to deal with the stubs, so it is really great that you have taken care of them.

Issue #173, opened 2020-08-22
Singular Matrix criteria is not correct (det, inv)


DeqingSun:
Hi, today I tried to do some perspective transformation, but ulab told me this matrix is singular. And no surprise, det of such matrix (N=4) is 0.

![Screen Shot 2020-08-21 at 11 02 52 PM](https://user-images.githubusercontent.com/1385273/90947313-da320480-e402-11ea-89ea-5d323fe79696.png)

I guess this triggered by:

https://github.com/v923z/micropython-ulab/blob/bbe39d9e41af5adb0292537777ea376813a6f7c8/code/linalg/linalg.c#L161
meanwhile linalg_invert_matrix uses similar code.

So I tried following code in colab and numpy:

```
m = np.array([[0,1], [1,0]])
print(np.linalg.det(m))
print(linalg.inv(m))
```
I got
```
-1.0
[[0. 1.]
 [1. 0.]]
```
But ulab gave me 0 and an Singular error.

v923z:
Thanks for taking action on this, I really appreciate it! The PR looks OK.

PR #175, opened 2020-08-23
Fix diagonal 0 error in row-reduction process

DeqingSun:
This fix https://github.com/v923z/micropython-ulab/issues/173
And the accuracy was pretty good compare to numpy.

### DIFF
--- a/code/linalg/linalg.c
+++ b/code/linalg/linalg.c
@@ -157,10 +157,27 @@ static mp_obj_t linalg_det(mp_obj_t oin) {
         tmp[i] = ndarray_get_float_value(in->array->items, in->array->typecode, i);
     }
     mp_float_t c;
+    mp_float_t det_sign = 1.0;
+
     for(size_t m=0; m < in->m-1; m++){
         if(MICROPY_FLOAT_C_FUN(fabs)(tmp[m*(in->n+1)]) < epsilon) {
-            m_del(mp_float_t, tmp, in->n*in->n);
-            return mp_obj_new_float(0.0);
+            //look for a line to swap
+            size_t m1=m+1;
+            for(; m1 < in->m; m1++){
+                if ( !(MICROPY_FLOAT_C_FUN(fabs)(tmp[m1*(in->n)+m]) < epsilon) ){
+                    for(size_t m2=0; m2 < in->n; m2++){
+                        mp_float_t swapVal = tmp[m*(in->n)+m2];
+                        tmp[m*(in->n)+m2] = tmp[m1*(in->n)+m2];
+                        tmp[m1*(in->n)+m2] = swapVal;
+                    }
+                    det_sign = -det_sign;
+                    break;
+                }
+            }
+            if (m1 >= in->m){
+                m_del(mp_float_t, tmp, in->n*in->n);
+                return mp_obj_new_float(0.0);
+            }
         }
         for(size_t n=0; n < in->n; n++){
             if(m != n) {
@@ -171,7 +188,7 @@ static mp_obj_t linalg_det(mp_obj_t oin) {
             }
         }
     }
-    mp_float_t det = 1.0;
+    mp_float_t det = det_sign;
 
     for(size_t m=0; m < in->m; m++){
         det *= tmp[m*(in->n+1)];
--- a/code/linalg/linalg.c
+++ b/code/linalg/linalg.c
@@ -55,8 +55,25 @@ bool linalg_invert_matrix(mp_float_t *data, size_t N) {
     for(size_t m=0; m < N; m++){
         // this could be faster with ((c < epsilon) && (c > -epsilon))
         if(MICROPY_FLOAT_C_FUN(fabs)(data[m*(N+1)]) < epsilon) {
-            m_del(mp_float_t, unit, N*N);
-            return false;
+            //look for a line to swap
+            size_t m1=m+1;
+            for(; m1 < N; m1++){
+                if ( !(MICROPY_FLOAT_C_FUN(fabs)(data[m1*N+m]) < epsilon) ){
+                    for(size_t m2=0; m2 < N; m2++){
+                        mp_float_t swapVal = data[m*N+m2];
+                        data[m*N+m2] = data[m1*N+m2];
+                        data[m1*N+m2] = swapVal;
+                        swapVal = unit[m*N+m2];
+                        unit[m*N+m2] = unit[m1*N+m2];
+                        unit[m1*N+m2] = swapVal;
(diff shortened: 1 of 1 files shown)


code/linalg/linalg.c
```diff
@@ -157,10 +174,27 @@ static mp_obj_t linalg_det(mp_obj_t oin) {
         tmp[i] = ndarray_get_float_value(in->array->items, in->array->typecode, i);
     }
     mp_float_t c;
+    mp_float_t det_sign = 1.0;
```
v923z, line 177: On some platforms, `mp_float_t det_sign = MICROPY_FLOAT_CONST(1.0)` is required. Could you, please, replace it?

code/linalg/linalg.c
```diff
@@ -55,8 +55,25 @@ bool linalg_invert_matrix(mp_float_t *data, size_t N) {
     for(size_t m=0; m < N; m++){
         // this could be faster with ((c < epsilon) && (c > -epsilon))
         if(MICROPY_FLOAT_C_FUN(fabs)(data[m*(N+1)]) < epsilon) {
-            m_del(mp_float_t, unit, N*N);
-            return false;
+            //look for a line to swap
```
v923z, line 58: Add your name on line 12!

v923z:
@DeqingSun Many thanks for looking into this! This could be merged, after you fix the two small issues I indicated in the comments.

code/linalg/linalg.c
DeqingSun: Will replace this one. I didn't know such Marco. After quick search it seems multiple files need to be changed as well?

```
$ grep -Er ' [0-9]+\.' .
./linalg/linalg.c:    mp_float_t elem = 1.0;
./linalg/linalg.c:	mp_float_t sum = 0.0;
./linalg/linalg.c:    mp_float_t det_sign = 1.0;
./linalg/linalg.c:        mp_float_t dot = 0.0;
./linalg/linalg.c:                mp_float_t sum = 0.0, v1, v2;
./linalg/linalg.c:        eigvectors[m*(in->n+1)] = 1.0;
./linalg/linalg.c:        largest = 0.0;
./linalg/linalg.c:    mp_float_t dot = 0.0, v;
./linalg/linalg.c:	mp_float_t trace = 0.0;
./ulab_create.c://|       Difference between consecutive elements, optional, defaults to 1.0
./ulab_create.c:        start = 0.0;
./ulab_create.c:        step = 1.0;
./ulab_create.c:        step = 1.0;
./ulab_create.c://|        and the other elements set to 0."""
./ulab_create.c://|    Return a new array of the given shape with all elements set to 1."""
./ulab_create.c://|    Return a new array of the given shape with all elements set to 0."""
./numerical/numerical.c:    mp_float_t value, sum = 0.0, sq_sum = 0.0;
./numerical/numerical.c:        sum = sq_sum = 0.0;
./numerical/numerical.c:        sum = 0.0;
./numerical/numerical.c:        sum_sq = 0.0;
./numerical/numerical.c:    // TODO: transpose the matrix, if axis == 0. Though, that is hard on the RAM...
./approx/approx.c://|     xtol: float = 2.4e-7,
./approx/approx.c://|     xatol: float = 2.4e-7,
./approx/approx.c://|     fatol: float = 2.4e-7,
./approx/approx.c:    //mp_float_t lambda = 0.0078125;
./approx/approx.c://|     xtol: float = 2.4e-7,
./approx/approx.c://|     rtol: float = 0.0,
./approx/approx.c://| def trapz(y: ulab.array, x: Optional[ulab.array] = None, dx: float = 1.0) -> float:
./ulab.h:// poly requires approx. 2.5 kB
./ulab.h:// the approx module consumes about 4.5 kB of flash space
./poly/poly.c:        XT[i+0*lenx] = 1.0; // top row
./poly/poly.c:            sum = 0.0;
./poly/poly.c:        sum = 0.0;
./poly/poly.c:        sum = 0.0;

```

DeqingSun: Also, is it necessary to change ```mp_obj_new_float(0.0);```? I guess so?

v923z: Yeah, that's correct.

v923z:
@jonathanhogg Jonathan, I was wondering, whether you could comment on this PR. You submitted https://github.com/v923z/micropython-ulab/pull/154, which addressed the problem of float constants. Deqing found a couple of other occurrences (see his long comment above), and I am not sure, how they slipped through last time. Since I don't see the problem that your PR fixed, I am a bit clueless, and could use a bit of advice here. Thanks!

jonathanhogg:
My understanding is that if you have code along the lines of:

```C
mp_float_t foo = 1.0;
```

then the compiler will be happy. On platforms where `mp_float_t` resolves to `float`, it will infer that `1.0` needs to be a `float` constant. However, if you have code like:

```C
mp_float_t foo;
foo = bar * 3.457;
```

It will take `3.457` as a `double` constant, compile a double-precision multiplication with `bar` and then cast it to a `float` for assignment to `foo`. This generates a compiler warning and, on the ESP32 port at least, the default build arguments interpret all warnings as errors and so the build fails.

So it's just the constants that are used in expressions that have to be wrapped in `MICROPY_FLOAT_CONST()`. There is, of course, no harm in just wrapping all float constants…

Either way, it'd be good to add an ESP32 platform build to the continuous-integration checks.

v923z:
> So it's just the constants that are used in expressions that have to be wrapped in `MICROPY_FLOAT_CONST()`. There is, of course, no harm in just wrapping all float constants…

Thanks for the thorough explanation!

> 
> Either way, it'd be good to add an ESP32 platform build to the continuous-integration checks.

I am not against the idea, but I don't work with the ESP32, so it would be a bit of hassle for me. However, if you are willing to update the workflow file, I would be happy to merge that.

code/linalg/linalg.c
DeqingSun: MICROPY_FLOAT_CONST added. At least linalg.c is good to go.

jonathanhogg:
> I am not against the idea, but I don't work with the ESP32, so it would be a bit of hassle for me. However, if you are willing to update the workflow file, I would be happy to merge that.

Figured you might say that 😉 – I'll add it to my to-do list…

v923z:
> Figured you might say that wink – I'll add it to my to-do list…

Thanks!

code/linalg/linalg.c
v923z: Thanks again for seeing this through till the end! You got your brownie points: https://github.com/v923z/micropython-ulab/graphs/contributors.

PR #182, opened 2020-10-23
Slice patch

v923z:
Fixes mysterious indexing error, when the index is a float.


kevinjwalters:
BTW, big Python and CircuitPython use `TypeError` for the wrong type for array index. `IndexError` seems to be used for things like out of bounds.

v923z:
@kevinjwalters Thanks for pointing this out!

PR #185, opened 2020-10-24
Fix doc build problem in CircuitPython

jepler:
CircuitPython doc build fails with this diagnostic:
```
/home/runner/work/circuitpython/circuitpython/shared-bindings/ulab/index.rst:220:'any' reference target not found: List[float]
```

Use double backticks instead of single backticks to differentiate between
a reference to a type and just "show in monospace font".

See https://github.com/adafruit/circuitpython/pull/3602 and https://github.com/adafruit/circuitpython/pull/3602/checks?check_run_id=1302834010


jepler:
This may be incomplete.  will continue working on it.

PR #186, opened 2020-10-24
Build circuitpython-ulab as part of CI

jepler:
This includes #185 so we can hopefully see if the problem is truly resolved.


jepler:
Hopefully this helps with maintaining CircuitPython compatibility!

v923z:
@jepler Thanks! In cases like this, you should really go ahead, and merge the PR. I am absolutely OK with that.

PR #220, opened 2020-11-16
add test script for array initialization

mouryarahul:
I've grouped almost all array creation methods into a single script. The outputs are visually cross-checked with Numpy.


tests/initialisation.py
```diff
@@ -0,0 +1,108 @@
...
+print(np.zeros((3,4), dtype=np.float))
+print("Array creation using ONES:")
+print(np.ones(3))
+print(np.ones((3,3)))
+print(np.ones((3,3), dtype=np.uint8))
+print(np.ones((3,3), dtype=np.uint16))
+print(np.ones((3,3), dtype=np.int8))
+print(np.ones((3,3), dtype=np.int16))
+print(np.ones((4,3), dtype=np.float))
+print(np.ones((3,4), dtype=np.float))
```
v923z, line 108: You have this line twice, and that is why the test crashes. You should replace 108 with a newline, and I think that would fix it.

tests/initialisation.py.exp
```diff
@@ -0,0 +1,188 @@
...
+array([[1, 1, 1],
+       [1, 1, 1],
+       [1, 1, 1]], dtype=int16)
+array([[1.0, 1.0, 1.0],
+       [1.0, 1.0, 1.0],
+       [1.0, 1.0, 1.0],
+       [1.0, 1.0, 1.0]], dtype=float64)
+array([[1.0, 1.0, 1.0, 1.0],
+       [1.0, 1.0, 1.0, 1.0],
+       [1.0, 1.0, 1.0, 1.0]], dtype=float64)
```
v923z, line 188: Perhaps, the newline character is missing here, too.

v923z:
@mouryarahul Thanks, this is a nice piece of test script. Before we merge it, however, we have to make sure that all checks (`micropython`, and `circuitpythoon`) pass. I believe, in this the problem is a single missing newline character.

tests/initialisation.py
mouryarahul: Thanks for the suggestion. Line 107 and 108 have a minor difference. I'll add the newline and resubmit.

v923z: That's right.

tests/initialisation.py.exp
mouryarahul: Yeah, I noticed that too.

v923z:
@mouryarahul Now all checks fail. Odd. I have to take off now, so if you can't sort it out, I can take it from here tomorrow.

mouryarahul:
> @mouryarahul Now all checks fail. Odd. I have to take off now, so if you can't sort it out, I can take it from here tomorrow.

I'm still new to this so can't figure out what's the problem. Maybe you can fix it and let me know. Thanks!

v923z:
@mouryarahul I am going to try to fix the first failure now, but you should create a new branch for each feature/test that you implement. That way it is much easier to keep track of where the error comes from.

mouryarahul:
> @mouryarahul I am going to try to fix the first failure now, but you should create a new branch for each feature/test that you implement. That way it is much easier to keep track of where the error comes from.

@v923z Thanks for the suggestions. I'll keep it in mind for next iteration.

v923z:
@mouryarahul These, and other, scripts fail already locally. I have the feeling that whenever `dtype=np....` is specified, that causes an error, and the script crashes. On a somewhat older version of `micropython` I don't see this problem, so something must have changed in `micropython` itself. The new executable is significantly larger (3.6 MB versus 600 kB).

mouryarahul:
> @mouryarahul These, and other, scripts fail already locally. I have the feeling that whenever `dtype=np....` is specified, that causes an error, and the script crashes. On a somewhat older version of `micropython` I don't see this problem, so something must have changed in `micropython` itself. The new executable is significantly larger (3.6 MB versus 600 kB).

@v923z Yes, I also find the same: specifying `'dtype=np....` causes the test to crash. But without specifying `dtype`, we can not perform extensive tests.
I do not totally understand what does the test process (run-tests.py) do, but I assume that it must be doing:
1. Run each script file and write the output as a file *.out
2. Compare the provided *.exp file with the *.out file using `diff` command and if there is no difference between the two, the test Passes otherwise it Fails.

When we run each test script on our terminal with `micropython` executable it produces the expected result and can write as a file using the command `>`. 
Will it be possible to write simpler Test process (script), which does the same, i.e., run each test script, write it's out as a file and then compare with the provided *.exp file? I'm not sure if this will work on GitHub too.

v923z:
> @v923z Yes, I also find the same: specifying `'dtype=np....` causes the test to crash. But without specifying `dtype`, we can not perform extensive tests.

This must be fixed, there is no question about that. 

> I do not totally understand what does the test process (run-tests.py) do, but I assume that it must be doing:
> 
>     1. Run each script file and write the output as a file *.out
> 
>     2. Compare the provided *.exp file with the *.out file using `diff` command and if there is no difference between the two, the test Passes otherwise it Fails.

That's pretty much it.

> Will it be possible to write simpler Test process (script), which does the same, i.e., run each test script, write it's out as a file and then compare with the provided *.exp file? I'm not sure if this will work on GitHub too.

github is just a linux computer, so that should be absolutely no problem. But the snag here is that whenever you have `dtype=...`, `micropython` crashes, so you won't get an output file. I will have to find out first, what changed in the handling of keyword arguments, so that the `ulab` code can be fixed. Patching up the test script is not an option in my opinion.

Sorry for all this trouble. I see that you have invested a lot of time into testing, and I definitely intend to merge your changes, but, as you see, unforeseen difficulties popped up.

When looking at your scripts, I also noticed quite a few things are awkward, because `scipy` stuff is mixed into `numpy`. Once the issue with testing is sorted out, I can re-arrange the code, so that it would look something like this

```python
from ulab import numpy as np
from ulab import scipy as spy
```

That would have the benefit of verbatim portability between computer and microcontroller. One problem there, however, is the question of what goes into the `vector` module. At the moment, that contains the vectorised versions of the `math` functions, but quite a few of them are actually part of `scipy.special` with a slightly different name.

jepler:
This change fixes the fatal micropython error `micropython: ../../py/argcheck.c:111: mp_arg_parse_all: Assertion '(allowed[i].flags & MP_ARG_KIND_MASK) == MP_ARG_OBJ' failed.`:
```
diff --git a/code/ndarray.c b/code/ndarray.c
index f4f7a38..84c6170 100644
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -795,7 +795,7 @@ ndarray_obj_t *ndarray_new_linear_array(size_t len, uint8_t dtype) {
 STATIC uint8_t ndarray_init_helper(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
     static const mp_arg_t allowed_args[] = {
         { MP_QSTR_, MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_rom_obj = mp_const_none } },
-        { MP_QSTR_dtype, MP_ARG_KW_ONLY, {.u_obj = MP_ROM_INT(NDARRAY_FLOAT) } },
+        { MP_QSTR_dtype, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_obj = MP_ROM_INT(NDARRAY_FLOAT) } },
     };
 
     mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
```

v923z:
@jepler Many thanks, Jeff! https://github.com/v923z/micropython-ulab/pull/235 This might upset `circuitpython`, I haven't tried it yet.

v923z:
@jepler Jeff, your code rectifies the problem in `micropython`, but we fail in `circuitpython`. Here is a minimal script that throws an error:

```python
try:
    import ulab as np
except ImportError:
    import numpy as np
    
print(np.array(range(2**8-5, 2**8), dtype=np.uint8))
print(np.array(range(2**8-5, 2**8), dtype=np.int8))
```
```bash
v923z:ulab$ ./micropython/ports/unix/micropython tests/init.py 
array([251, 252, 253, 254, 255], dtype=uint8)
array([-5, -4, -3, -2, -1], dtype=int8)
```
is OK, but 
```bash
v923z:ulab$ ./circuitpython/ports/unix/micropython tests/init.py 
array([251, 252, 253, 254, 255], dtype=uint8)
  File "tests/init.py", line 7, in <module>
OverflowError: value must fit in 1 byte(s)
```
doesn't look so good. It seems to me that we are caught in the `ndarray_make_new_core` function 
https://github.com/v923z/micropython-ulab/blob/31d0bec9fe58c5170865a7f77609cd89296b6eed/code/ndarray.c#L868-L871
I am a bit surprised, because I thought that this is so deep in the core that it wouldn't have diverged from `micropython`. I can try to get to the root of the problem.

v923z:
@mouryarahul Rahul, we might have to re-arrange the test scripts a bit, because we can't possibly test against `micropython` and `circuitpython` with the same set of scripts. With the latest changes, your `initialisation.py` passes on `micropython`, but fails on `circuitpython`, and the error is quite fundamental, we can't fix it in `ulab`. In principle, we could merge your tests now, but I would really like to have a clean state. 

I will try to pull this off tomorrow. Sorry for the delay!

mouryarahul:
> @mouryarahul Rahul, we might have to re-arrange the test scripts a bit, because we can't possibly test against `micropython` and `circuitpython` with the same set of scripts. With the latest changes, your `initialisation.py` passes on `micropython`, but fails on `circuitpython`, and the error is quite fundamental, we can't fix it in `ulab`. In principle, we could merge your tests now, but I would really like to have a clean state.
> 
> I will try to pull this off tomorrow. Sorry for the delay!

@v923z No worries. If required some changes in the test scripts, I'll do them. Let me know your suggestions. I've found few glitches in the test script results (*.exp files). I'll correct them and do the PR again.

v923z:
@mouryarahul I think you should no longer have difficulties, but you'll have to move your scripts: https://github.com/v923z/micropython-ulab/pull/237

mouryarahul:
Will do a new PR.

PR #230, opened 2020-11-18
pin micropython 1.13 for tests

jepler:


jepler:
Note that by default ./build.sh uses an existing clone of micropython, so locally it will be necessary to `rm -rf micropython` before running this `./build.sh` to see the results of the change.

PR #237, opened 2020-11-20
split test scripts into numpy-compatible and circuitpy-compatible sets

v923z:
This PR implements independent testing of `numpy`-compatible, and `circuitpy`-compatible features. Test files that are independent of the platform should be in `tests/common`, while test scripts that can run on a specific platform only should be placed into `tests/numpy`, and `tests/circuitpy`, respectively. 

This should fix the issue encountered in https://github.com/v923z/micropython-ulab/pull/220


v923z:
@jepler Jeff, many thanks for all your efforts, code hacking, advisory services, chits and chats and the like!

v923z:
@jepler Jeff, many thanks for all your efforts, code hacking, consulting services, chits and chats and the like!

PR #243, opened 2020-11-20
add script to test linalg module

mouryarahul:
This branch contains the script to test linalg module:
1. ulab.linalg.dot

1. ulab.linalg.inv

1. ulab.linalg.det

1. ulab.linalg.eig

1. ulab.linalg.cholesky

1. ulab.linalg.norm

1. ulab.linalg.trace


tests/numpy/linear_algebra.py
```diff
@@ -0,0 +1,113 @@
...
+    b = np.linalg.inv(a)
+    ab = np.linalg.dot(a, b)
+    m,n = ab.shape()
+    for i in range(m):
+        for j in range(n):
+            if i == j:
+                print(math.isclose(ab[i][j], 1.0, rel_tol=1E-9, abs_tol=1E-9))
+            else:
+                print(math.isclose(ab[i][j], 0.0, rel_tol=1E-9, abs_tol=1E-9))
+else:
```
v923z, line 42: Since the only difference of the two code blocks is the use of `ab.shape()` vs. `ab.shape`, I would solve it like so (untested)

```python
def tensor_shape(tensor):
    if try:
        return tensor.shape
    except:
        return tensor.shape()
```
That would reduce code redundancy by a huge factor. It would also mean that in the future we would have to replace a single line of code. I have a pending PR in `micropython`, you can support it there: https://github.com/micropython/micropython/pull/5648

mouryarahul: > ```python
> def tensor_shape(tensor):
>     if try:
>         return tensor.shape
>     except:
>         return tensor.shape()
> ```

@v923z It's a good suggestion, but the differences between the two blocks are not only the `shape`, but also the `dot` method: in Numpy, it's `np.dot` where as in ULAB, it's `ulab.linalg.dot`. So, I think we could keep the two blocks until we make changes in ULAB to make it `ulab.dot.` I will look into the PR to support it if I understand it correctly.

v923z: This is indeed so, but then we would have to take `.dot` out of `linalg`, and move it to `numerical`, perhaps. Somehow, I overlooked this; it `.dot` looks like a linear algebra function. Let's leave it for now.

PR #247, opened 2020-11-21
add script to test polynomial module

mouryarahul:
This branch contains the script to test polynomial fitting functions:

1. ulab.polyval

1. ulab.polyfit


v923z:
@mouryarahul Rahul, could you, please, rename this as `poly.py`, because `polyval.py` is misleading.

mouryarahul:
> @mouryarahul Rahul, could you, please, rename this as `poly.py`, because `polyval.py` is misleading.

@v923z in `tests\numpy` there is an existing file `poly.py` and `poly.py.exp`. So,  I will rename the `polyva.py` to `poly_extra.py`.

v923z:
> > @mouryarahul Rahul, could you, please, rename this as `poly.py`, because `polyval.py` is misleading.
> 
> @v923z in `tests\numpy` there is an existing file `poly.py` and `poly.py.exp`. So, I will rename the `polyva.py` to `poly_extra.py`.

You should really either combine the two scripts, or get rid of the old one. It will quickly become unmanagable, if we have tests for one function in two places.

mouryarahul:
> > > @mouryarahul Rahul, could you, please, rename this as `poly.py`, because `polyval.py` is misleading.
> > 
> > 
> > @v923z in `tests\numpy` there is an existing file `poly.py` and `poly.py.exp`. So, I will rename the `polyva.py` to `poly_extra.py`.
> 
> You should really either combine the two scripts, or get rid of the old one. It will quickly become unmanagable, if we have tests for one function in two places.

@v923z `poly_extra` includes all the tests in `poly`, so I'll delete the `poly`.

Issue #250, opened 2020-11-21
arrange functions into truly `numpy`-compatible structures


v923z:
Functions can now be called in a `numpy`-compatible fashion, or, following `circuitpython`'s convention, importing sub-modules, by appropriately setting the `ULAB_NUMPY_COMPATIBILITY` constant in https://github.com/v923z/micropython-ulab/blob/5f7092cec0fb79db4446b72ab3dca98ab83fb53e/code/ulab.h#L31

However, some of the functions are actually `scipy`-related. In order to make porting existing code, `ulab` should be split into two sub-modules, each entirely compatible with existing `numpy` and `scipy` conventions. `ulab` could then be used as 

```python
from ulab import numpy as np
from ulab import scipy as spy


a = np.sin(...)
b = np.fft.fft(a)

c = spy.optimize(...)
```
etc.

v923z:
Can be closed, since v.2.1 implements the `numpy`-compatible interface.

PR #263, opened 2020-11-23
for discussion: tests: try to fit mp, cp, and py3 into one mold

jepler:
Mostly by providing a "numpy shim" for CircuitPython,
try to make the numpy tests run on all three systems.
(a "scipy shim" might also be useful?)

However, there are test failures.  Is it worth working
through them and getting this to a point where it could
be included?


v923z:
> However, there are test failures. Is it worth working
> through them and getting this to a point where it could
> be included?

I think an implementation of https://github.com/v923z/micropython-ulab/issues/250 would significantly reduce the complexity of such a compatibility layer.

PR #270, opened 2020-12-21
Add `array.dtype` stub

### LINKED ISSUE #269: `array.dtype` stub & docs
Hello and thank you very much for your work on this project.

I think the `array.dtype` stub is missing from the pyi comments.
I've added it in locally so that I can continue to rely on the stubs, and will submit a pull request shortly.

Also, I think that the current `latest` docs for dtype are incorrect as it is stated to be a property whereas currently on master it is a regular method. I believe the docs are desired behaviour, mirroring the numpy api. This might be being worked on in [the numpy-scipy branch](https://github.com/v923z/micropython-ulab/blob/numpy-scipy) but I figured I'd ask just to be sure.

CallumJHays:
Attempts to fix #269

### DIFF
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -100,6 +100,10 @@ mp_uint_t ndarray_print_edgeitems = NDARRAY_PRINT_EDGEITEMS;
 //|         """Return a copy of the array"""
 //|         ...
 //|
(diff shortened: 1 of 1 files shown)


v923z:
@CallumJHays Fantastic, thanks a lot!

PR #286, opened 2021-01-13
Fix macos build (mp & cp), add mac builds to ci

jepler:


v923z:
@jepler Thanks, Jeff!

PR #288, opened 2021-01-15
frombuffer: Add type declarations to stubs

jepler:
Without this, circuitpython doesn't succeed at `make check-stubs` during its build process:
```
circuitpython-stubs/ulab/__init__.pyi:320: error: Function is missing a type annotation for one or more arguments
```


v923z:
Oh, snap, I have forgotten that. Thanks for catching it!

PR #293, opened 2021-01-23
fix to allow including user.c

teuler:
Solves the issue that constants (e.g. `ULAB_USER_MODULE` -> `ULAB_HAS_USER_MODULE`) have changed, preventing to include `/user/user.c`.


v923z:
@teuler Thanks!

Issue #274, opened 2021-01-08
Add numpy.isinf() & numpy.isfinite()


CallumJHays:
`np.isinf` https://numpy.org/doc/stable/reference/generated/numpy.isinf.html

`np.isfinite` https://numpy.org/doc/stable/reference/generated/numpy.isinf.html

v923z:
These two functions should be added to https://github.com/v923z/micropython-ulab/blob/master/code/compare/compare.c

v923z:
@CallumJHays Cal, would https://github.com/v923z/micropython-ulab/tree/isinf do? I have also added some documentation here: https://github.com/v923z/micropython-ulab/blob/isinf/docs/manual/source/numpy-functions.rst. 

It would be really great, if you could add some tests in https://github.com/v923z/micropython-ulab/tree/isinf/tests. If you drop your test script under `tests/numpy`, and run the `build.sh` script in the root, that should automatically produce the expected results. Let me know if you need help with that.

CallumJHays:
@v923z thanks! I was just about to get onto these. I'll try to get them done tomorrow

PR #300, opened 2021-02-01
Add `np.isinf` tests

CallumJHays:
I pretty much just copied the examples from your documentation for the tests and added one more example for extremely large float literals. Hopefully this wraps up #274 , I couldn't think of what else to add at the time being.


v923z:
Thanks, this should do!

PR #314, opened 2021-02-10
Add vector 2-norm test

ubIQio:
Fails on high tolerance if result is >10?

### DIFF
--- a/tests/numpy/linalg.py
+++ b/tests/numpy/linalg.py
@@ -106,19 +106,18 @@ result = (np.linalg.norm(a))  ## Here is a problem
 ref_result = 16.881943016134134
 print(math.isclose(result, ref_result, rel_tol=1E-6, abs_tol=1E-6))
 
-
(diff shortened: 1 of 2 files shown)


v923z:
@ubIQio I wouldn't get too hung up on the tolerance issue: some platforms have only floats. I would say, if you look at the results from `numpy`, and from `ulab`, and they are reasonably close, then I would accept them. 

However, you have to update the expected file https://github.com/v923z/micropython-ulab/blob/master/tests/numpy/linalg.py.exp. The easiest way of doing that is to simply run the https://github.com/v923z/micropython-ulab/blob/master/build.sh script. That pulls in the latest `micropython`, compiles the firmware for `unix`, and runs the tests. At the end, if a test fails, or its expected file is not in `./tests/`, then a new expected file will be generated in the root. You'll have to copy that to `./tests/numpy/` as `mytest.py.exp`. At that point, `mytest.py.exp` becomes the gold standard for `mytest.py`, and the results of `mytest.py` will in the future be compared against this file.

v923z:
@ubIQio This looks good, many thanks for your efforts!

PR #327, opened 2021-02-18
Dtype mod

v923z:
This PR adds the option to extend `ulab` in a transparent way. This means that the user is able to add their own data container in the C implementation, and if they supply a readout function, then various `numpy` methods should be able to access the data in the container. Such a facility could be exploited to process data that do not reside in RAM, either because they are not available, or because the amount would be prohibitive.

Two possible use cases are

1. implementing complicated generator expressions
2. processing image data that contain megapixels of information (openmv, https://github.com/openmv/openmv/issues/881)

The `dtype` member of the `ndarray` is extended as 

```c
typedef struct _dtype_dtype {
    uint8_t type;
    #if ULAB_DTYPE_IS_EXTENDABLE
    uint8_t flags;
    void *arrfunc;
    uint8_t *subarray;
    size_t shape[ULAB_MAX_DIMS];
    void *origin;
    uint8_t *name;
    #endif
} dtype_dtype;
```
where a pointer to the readout function can be attached, `*arrfunc`, as well as a temporary container, `*subarray`, can be pointed to. The `subarray` has to be able to hold a single line of data. This single line can than be passed to the innermost loop of all numerical functions, binary operators, etc. An example is the summation macro

```c
#define RUN_SUM1(type, ndarray, array, results, rarray, ss, arrfunc)\
({\
    type sum = 0;\
    int32_t increment = (ss).strides[0];\
    (ndarray)->dtype.subarray = (array);\
    if((ndarray)->dtype.flags) {\
        (arrfunc)((ndarray), (array), &increment, (ss).shape[0]);\
    }\
    for(size_t i=0; i < (ss).shape[0]; i++) {\
        sum += *((type *)((ndarray)->dtype.subarray));\
        (array) += increment;\
    }\
    (array) += (ss).shape[0] * (ss).strides[0];\
    memcpy((rarray), &sum, (results)->itemsize);\
    (rarray) += (results)->itemsize;\
})
```

The user can then simply attach their readout function 

```c
static void user_imreader(ndarray_obj_t *ndarray, void *array, int32_t *strides, size_t count) {
    uint16_t *subarray = (uint16_t *)ndarray->dtype.subarray;
    // if necessary, get the coordinates in the original reference frame, i.e.,
    // in the coordinates used at the time of the creation of the object
    // size_t *coords = tools_coords_from_pointer(array, ndarray);
    for(size_t i = 0; i < count; i += *strides/ndarray->itemsize) {
        // fill up the array with dummy data
        *subarray++ = i*i;
    }
    // since strides is going to be used in computation loops, and subarray is
    // meant to be a dense array, simply overwrite strides with the itemsize
    *strides = ndarray->itemsize;
}

static mp_obj_t user_imread(mp_obj_t shape) {

    size_t len = mp_obj_get_int(shape);

    ndarray_obj_t *ndarray = ndarray_new_linear_array(len, NDARRAY_UINT16);
    uint16_t *array = (uint16_t *)ndarray->array;
    for(size_t i = 0; i < len; i++) {
        *array = (uint16_t)i;
        array += 1;
    }
    ndarray->dtype.flags = 1;
    ndarray->dtype.arrfunc = user_imreader;
    uint8_t *subarray = m_new(uint8_t, ndarray->itemsize * ndarray->shape[ULAB_MAX_DIMS - 1]);
    ndarray->dtype.subarray = subarray;
    memcpy(&(ndarray->dtype.shape), &(ndarray->shape), sizeof(size_t) *ULAB_MAX_DIMS);
    return MP_OBJ_FROM_PTR(ndarray);
}
```


kwagyeman:
This is more or less what we need to support images. The data type we'd use would be a float. Since users have often asked to do weird things with images that you can't do with ints.

What questions do you need answered from me? Given a row/color_channel it's very easy for me to fill a float array with pixel values. I can also easily fill a column.

v923z:
@kwagyeman 

> This is more or less what we need to support images. The data type we'd use would be a float. Since users have often asked to do weird things with images that you can't do with ints.

I don't have to know what data type you want to hold, `int`, or `float`. This will be resolved automatically, when the function pointer is called. 

> 
> What questions do you need answered from me? Given a row/color_channel it's very easy for me to fill a float array with pixel values. I can also easily fill a column.

I think, we are pretty much on the same page. Let me iron out the implementation, and we can then pick it up from there.

v923z:
@kwagyeman @iabdalkader What should happen with this construct, if the one wants to use the buffer protocol? Some context can be found here https://github.com/v923z/micropython-ulab/issues/335, and here https://github.com/v923z/micropython-ulab/issues/328,  

The problem is that in https://github.com/v923z/micropython-ulab/blob/42212622ff3c49b5f03907aac8788a759e550ba0/code/ndarray.c#L2031-L2040 we have to set a pointer to the underlying data, which I won't hold, because `self->array` will not point to actual data. 

I think the cleanest solution is to simply bail out, if the special flag is set for an `ndarray`. Do you agree? But you still have to point this out in your documentation.

kwagyeman:
We have a buffer protocol for the image object already. There is no need to duplicate it. So bailing makes sense.

v923z:
> We have a buffer protocol for the image object already. There is no need to duplicate it. So bailing makes sense.

OK, thanks!

kwagyeman:
@v923z Would you like to get on our slack? Email kwagyeman@openmv.io

v923z:
> @v923z Would you like to get on our slack? Email [kwagyeman@openmv.io](mailto:kwagyeman@openmv.io)

Thanks for the invitation! Sure!

kwagyeman:
You need to email me since you hide all contact info on your public profile.

v923z:
@kwagyeman, @iabdalkader I have updated my original comment, and uploaded a working prototype. At the moment, only `sum`, `std`, and `mean` are supported, and you can iterate over the tensor elements. Adding the rest is not hard, but we should first converge on an interface function.  The example 

```from ulab import blocks
from ulab import user
from ulab import numpy as np

f = blocks.ndarray(shape=(5,5), transformer=user.imreader(), dtype=np.uint8)

print(f)
print(np.sum(f, axis=1))
for i in f:
    print(i, np.sum(i, axis=0))
```
creates an `ndarray` header with `shape=(5,5)`, attaches the function pointer via `user.imreader`, and sets the `dtype` to `uint8`. From this point, `f` behaves like an `ndarray`, except, it fetches the data by means of the function pointer, if data are needed. I simply return square numbers in https://github.com/v923z/micropython-ulab/blob/3227831a0adfc70c090664d8c8f9ae212a9a220a/code/user/user.c#L95-L98. This is, where one would implement a function that reads actual pixels. 

I think this solution is quite flexible (by which I mean that the `ulab` core doesn't have to know anything about the implementation of `imreader`, it can completely be detached), but I might very well have overlooked something. Let me know what you think. 

What would, perhaps, be great is, if we could call the `.print` function of `imreader`, so that in https://github.com/v923z/micropython-ulab/blob/3227831a0adfc70c090664d8c8f9ae212a9a220a/code/blocks/blocks.c#L51-L65, we could also indicate, which `transformer` is used in a particular block. I haven't yet found a way of doing this, however.

dhalbert:
@v923z This is very interesting. I am unsure: is this only for reading out-of-RAM data? Do you need a corresponding write function?

v923z:
@dhalbert 

> is this only for reading out-of-RAM data? Do you need a corresponding write function?

At the moment, this would only read; we could think about adding a write function.

PR #329, opened 2021-02-19
ndarray: Fix memoryview(ulab.array(...))

### LINKED ISSUE #328: 1.7.2: ndarray_get_buffer() thinks `->array` is a python object
ulab 1.7.2 still defines the routine `ndarray_get_buffer()`. This code calls `mp_get_buffer(self->array, ...)`. However, my understanding from reading `create_zeros_ones_full()` is that `self->array` is the raw array of values. It is not a Python object. `mp_get_buffer()` expects its first arg to be a Python object, so bad things happen.

https://github.com/v923z/micropython-ulab/blob/743d86487c83e42024ed508ed50499ad0a527d5d/code/ndarray.c#L1989-L1994

```c
mp_int_t ndarray_get_buffer(mp_obj_t self_in, mp_buffer_info_t *bufinfo, mp_uint_t flags) {
    ndarray_obj_t *self = MP_OBJ_TO_PTR(self_in);
    // buffer_p.get_buffer() returns zero for success, while mp_get_buffer returns true for success
    return !mp_get_buffer(self->array, bufinfo, flags);
}
```

I had some code similar to this in CircuitPython:
```python
a = ulab.zeros(10, dtype=ulab.uint16)
some_bleio_characteristic.value = a
```
`_bleio.Characteristic.value` expects the value coming in to obey buffer protocol, and calls `mp_get_buffer()` on the `ndarray`. That in turn calls `ndarray_get_buffer()`, which tries to call `mp_get_buffer()` on the `array` itself, and things go wrong.

`ndarray_get_buffer()` seems to have disappeared in 2.x `ulab`, so I'm not sure how to fix 1.7.2. Is there something new to backport to 1.7.x? Thanks.

jepler:
For now this only handles the 1D case.  In theory it would work for any dense array, however, I found that `ndarray_is_dense` didn't behave for me so I implemented this instead.

Add a test.  Before the change, this test would segfault.

Closes #328.


code/ndarray.c
```diff
@@ -1986,8 +1986,15 @@ mp_obj_t ndarray_info(mp_obj_t obj_in) {
 MP_DEFINE_CONST_FUN_OBJ_1(ndarray_info_obj, ndarray_info);
 #endif
 
+// (the get_buffer protocol returns 0 for success, 1 for failure)
 mp_int_t ndarray_get_buffer(mp_obj_t self_in, mp_buffer_info_t *bufinfo, mp_uint_t flags) {
     ndarray_obj_t *self = MP_OBJ_TO_PTR(self_in);
-    // buffer_p.get_buffer() returns zero for success, while mp_get_buffer returns true for success
-    return !mp_get_buffer(self->array, bufinfo, flags);
+    if (self->ndim != 1 || self->strides[0] > 1) {
+        // For now, only allow fetching buffer of a 1d-array
```
v923z, line 1993: This should be OK, though, while we can't specify the shape anyway. Once https://github.com/v923z/micropython-ulab/pull/305 is done, one could initialise from a buffer with `shape` and `strides`.

v923z:
@jepler Many thanks! I haven't even had time to look at @dhalbert's bug report, before the fix arrived.

v923z:
>  however, I found that `ndarray_is_dense` didn't behave for me so I implemented this instead.

Do you mean that the function is corrupt? I might have to look into that, then.

PR #330, opened 2021-02-19
ndarray: Use ndarray_is_dense() to decide if array is buffer-able

jepler:
This doesn't work as I expected, because `np.eye(4)` isn't reported as 'dense'.  If I understand correctly, a "dense" array is one where the storage of all the elements is contiguous, without gaps.  Perhaps it is my misunderstanding, or perhaps it's expected that `np.eye()` is not dense in general.


v923z:
> This doesn't work as I expected, because `np.eye(4)` isn't reported as 'dense'. If I understand correctly, a "dense" array is one where the storage of all the elements is contiguous, without gaps. Perhaps it is my misunderstanding, or perhaps it's expected that `np.eye()` is not dense in general.

@jepler I think your expectation is correct, however, the implementation is faulty. The problem is this: 

https://github.com/v923z/micropython-ulab/blob/8d156616bf0ad1993e76357612623f59fdc9153c/code/ndarray.c#L648-L649

The `i = ULAB_MAX_DIMS` index accesses a RAM segment that is out of bounds, and was never initialised. 

Here is the fix. https://github.com/v923z/micropython-ulab/commit/f4604d208b9d92d4fa8869b74a54ebb71a6ed101 Accidentally, I pushed it directly into the `legacy` branch, sorry!

jepler:
@v923z I think something's still weird, when I try to use this patch rebased onto legacy on a non-square array:
```python
>>> import ulab as mp
>>> list(memoryview(np.ones((3,2), dtype=np.int8)))
TypeError: object with buffer protocol required
>>> list(memoryview(np.ones((3,3), dtype=np.bool)))
[1, 1, 1, 1, 1, 1, 1, 1, 1]
>>> list(memoryview(np.ones((2,3), dtype=np.bool)))
TypeError: object with buffer protocol required
```
It did fix _square_ arrays.

jepler:
Maybe it doesn't matter for memoryview though, for python3 compat:
```
>>> import numpy as np
>>> list(memoryview(np.ones((3,3))))
NotImplementedError: multi-dimensional sub-views are not implemented
```

jepler:
and, not that anybody cares, on python2.7 this consumed at least 1GB of memory before I stopped it.  No idea why, and they won't take my bug reports about python2.7 anymore
```
>>> import numpy as np
>>> list(memoryview(np.ones((3,3))))
```

v923z:
@jepler I was probably lazy to clone your repository, so I created a new branch. It turns out, I went too far to the left with the indexing. The fix is in https://github.com/v923z/micropython-ulab/pull/334.

```python
import ulab as np

a = np.ones((3,2), dtype=np.int8)
print(list(memoryview(a)))
```
```
[1, 1, 1, 1, 1, 1]
```
However, it seems to me that the `ndarray_is_dense` function is called twice in the snippet above, and I don't know, where exactly. Does the statement `memoryview(a)` have to pass twice somewhere in `micropython`?

Issue #341, opened 2021-03-03
Support for dtypes uint32 and int32


pumelo:
**Describe the solution you'd like**
Support dtypes int32 and uint32.

**Additional context**
I do get an array of int32 from other hardware (i2s microphone). I did currently not find an efficient way to use this data with ulab than iterating over the array and converting it to an array of floats. This seems cumbersome and slow. So it would be nice to have int32 directly supported.

This is what I am doing right now:

```
samples = bytearray(2048)  # bytearray to receive audio samples
data = uctypes.struct(uctypes.addressof(samples), {'arr': (0 | uctypes.ARRAY, 512 | uctypes.INT32)})

np_a = np.zeros(512, dtype=np.float)

# <--  here, get data from hardware into samples buffer

for i in range(512):
    np_a[i] = data.arr[i]
```

This loop above loop takes about 90ms on a esp32s2 @ 240MHz which is insanely slow. While the fft on the same data thereafter takes about 10ms.

pumelo:
Just found an "easy" way to get a performance boost (overally I'd still consider this slow 😸):

putting the conversion into its one function speeds the convertion up by approximatly a factor of two. I think this is mostly due to the attribute access and lookup which is not present. (data.arr[i] vs. input[i]). I think this is mentioned somewhere in the micropython docs.. 

```
samples = bytearray(2048)  # bytearray to receive audio samples
data = uctypes.struct(uctypes.addressof(samples), {'arr': (0 | uctypes.ARRAY, 512 | uctypes.INT32)})

np_a = np.zeros(512, dtype=np.float)

def i32tof32(input, output):
    for i in range(512):
        output[i] = input[i]

# <--  here, get data from hardware into samples buffer

i32tof32(data.arr, np_a)
```

v923z:
@pumelo 

> Support dtypes int32 and uint32.

I think the short answer is no. If you want to know the reason, you can read more about the problems here: https://github.com/v923z/micropython-ulab/issues/306#issuecomment-774664893

However, if I understand your problem correctly, you do not want complete support, you only want to pass an `int32` array into a `numpy` function. This is a subtle, but significant difference. 

[...]

Issue #306, opened 2021-02-06
frombuffer byte order


ubIQio:
I have a buffer of data from an ADC, which i am trying to access as an ndarray

The 1024 byte buffer is organised in 'frames' of 8 bytes, the bytes of interest in each frame  being 1 through 6, being three pairs of 16 bit samples.

If I ignore the last byte of the buffer, I can get the buffer as an ndarray uising

``` python
buffer = b'\x40\x00\xda\xff\xd6\xc1\x2e\x0d\x40\x00\xde\xff\xca\xc1\x3c\x0d\x40\x00\xe6\xff\xc4\xc1\x40\x0d\x40\x00\xe0\xff\xb8\xc1\x32'


a = np.frombuffer(buffer, dtype=np.uint16, offset=1)
```
but the result is byteswapped values e.g

```
a:	 array([55808, 55039, 11969, ..., 57344, 47359, 12993], dtype=uint16)
```

dtype = np.>uint16 causes a reboot, any ideas how this can be achieved?

v923z:
> I have a buffer of data from an ADC, which i am trying to access as an ndarray
> 
> The 1024 byte buffer is organised in 'frames' of 8 bytes, the bytes of interest in each frame being 1 through 6, being three pairs of 16 bit samples.
> 
> If I ignore the last byte of the buffer, I can get the buffer as an ndarray uising
> 
> ```python
> buffer = b'\x40\x00\xda\xff\xd6\xc1\x2e\x0d\x40\x00\xde\xff\xca\xc1\x3c\x0d\x40\x00\xe6\xff\xc4\xc1\x40\x0d\x40\x00\xe0\xff\xb8\xc1\x32'
> 
> 
> a = np.frombuffer(buffer, dtype=np.uint16, offset=1)
> ```
> 

> but the result is byteswapped values e.g
> 
> ```
> a:	 array([55808, 55039, 11969, ..., 57344, 47359, 12993], dtype=uint16)
> ```

I am not sure, I see the problem here. First, this is the `numpy` behaviour: 

```buffer = b'\x40\x00\xda\xff\xd6\xc1\x2e\x0d\x40\x00\xde\xff\xca\xc1\x3c\x0d\x40\x00\xe6\xff\xc4\xc1\x40\x0d\x40\x00\xe0\xff\xb8\xc1\x32'
a = np.frombuffer(buffer, dtype=np.uint16, offset=1)

[...]

PR #342, opened 2021-03-03
added first implementation of utils sub-module

v923z:
This PR should address the issue raised in https://github.com/v923z/micropython-ulab/issues/341

The general problem is that peripheral devices do not necessarily supply data compatible with `ulab`'s 5 supported `dtypes`. However, for the most common tasks, simple conversion functions can be added. With the help of these, arbitrary data formats could be converted to the supporetd `dtype`s. After the conversion, the standard `ulab` operations can be carried out on the data arrays.

In the future, the `utils` sub-module could be the place, where useful, but not `numpy`-compatible functions working with or returning `ndarray`s could be stored.


code/utils/utils.c
```diff
@@ -0,0 +1,137 @@
...
+                    memcpy(--tmpbuff, buffer++, 1);
+                }
+                if(buffer_type == UTILS_INT32_BUFFER) {
+                    *array++ = (mp_float_t)(*(int32_t *)tmpbuff);
+                } else {
+                    *array++ = (mp_float_t)(*(uint32_t *)tmpbuff);
+                }
+            }
+        } else {
+            for(uint8_t i = 0; i < len; i++) {
```
pumelo, line 92: use size_t? uint8_t will probably not reach the end condition...

pumelo (changes requested):
This is probably the cause for the exception I saw.. About to recompile and test .. 🏃

code/utils/utils.c
pumelo: Yes, this was the issue! with the documented crash.

v923z: Indeed! 😊 Thanks for catching this!

v923z:
@ubIQio  This can also solve your issue https://github.com/v923z/micropython-ulab/issues/306

PR #347, opened 2021-03-10
:pencil: Add rp2 firmware instructions

jsimonrichard:
I added instructions for compiling the firmware on the rp2 port (as requested).


v923z:
@jsimonrichard Thanks for your contribution! Much appreciated.

PR #356, opened 2021-03-30
Update RP2 build script and README instructions

Gadgetoid:
The `USER_C_MODULE` support for `cmake`-based ports is now upstream :sparkle: and this PR should update the documentation and build script to reflect this!


v923z:
@Gadgetoid Fantastic, many thanks!

PR #359, opened 2021-04-01
code: Add a docstring for numpy, scipy packages

jepler:
.. this fixes a problem in CircuitPython where building the docs would say "../shared-bindings/ulab/numpy/approx/index.rst: WARNING: document isn't included in any toctree"

### DIFF
--- a/code/numpy/approx/approx.c
+++ b/code/numpy/approx/approx.c
@@ -28,17 +28,17 @@ const mp_obj_float_t approx_trapz_dx = {{&mp_type_float}, MICROPY_FLOAT_CONST(1.
 
 #if ULAB_NUMPY_HAS_INTERP
 //| def interp(
-//|     x: ulab.array,
-//|     xp: ulab.array,
-//|     fp: ulab.array,
+//|     x: ulab.ndarray,
+//|     xp: ulab.ndarray,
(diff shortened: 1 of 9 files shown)


v923z:
@jepler Jeff, thanks for looking into this, I have slipped back with the python stubs. 

I was wondering, why you were basing the PR on 2.1.5, and then I realised that that is the last release. I have been waiting for the fix in https://github.com/v923z/micropython-ulab/pull/305, but I understand that you want to have a proper anchor, so I have just generated a new tag https://github.com/v923z/micropython-ulab/releases/tag/2.6.0. Could you, please, update the PR with this? That should resolve the conflict that is caused by my having moved a function from `linalg`.

However, the question of the python stubs leads to another problem, namely that the documentation of the functions is scattered now. Earlier, this wasn't an issue, because the sub-modules were relatively small, so each of them had a single source file. This is no longer feasible, with `numpy` having close to 50 functions. 

I am not quite sure what the proper solution is. We could separate the stubs into a single file, but you probably wouldn't like that. (I am no fan, either.) The other option is that the script that produces the documentation should contain the document structure. I believe the "just take all files from these directories and process them one by one" approach won't work anymore.

PR #360, opened 2021-04-03
Changes relative to 2.6.0 to allow building on CircuitPython

jepler:


v923z:
@jepler Jeff, thanks for all the hard work! Looking at the changes, this was actually more than I had expected. I don't see any issues with the PR, so you can either merge it, or extend it, as you wish.

Issue #364, opened 2021-04-14
[FEATURE REQUEST] support for complex dtype


v923z:
I would like to discuss the ramifications of adding support for complex arrays in `ulab`. 

At the beginning, the Fourier transform was the single function that would have required complex arrays, but it was possible to just pass and return two real arrays that stood for the real and imaginary parts. This was not very elegant, but still acceptable. However, there have been at least two feature requests recently that would result in functions that lead out of the domain of real numbers. 

Namely, 

1. eigenvalues/eigenvectors of generic matrices https://github.com/v923z/micropython-ulab/issues/363
2. roots of polynomials https://github.com/v923z/micropython-ulab/issues/355

The problem is that these functions could potentially return complex arrays, even if the input was real. 

Adding a new `dtype` increases the firmware size, because the `dtype`s have to be resolved in C; in particular, the biggest contributor is probably the set of binary operators, because there one has to tackle all possible combinations, therefore, this contribution to the firmware size scales with the square of the number of `dtypes`. At the moment, we have 5 and a half `dtypes` (`(u)int8`, `(u)int16`, `float/double`, plus `bool`, which is actually an `uint8` with a special flag), hence we deal with 25 combinations. By the addition of `complex`, we would end up with 36 combinations. 

There are a number of functions that would not have to support `complex`. E.g., `(arg)min`, `(arg)max`, `(arg)sort`, `median`, `clip`, `fmin` etc. With these considerations, a rough estimate is 30% extra in the firmware size, if we add `complex` as a `dtype`.

Before jumping off the deep end, here are a couple of questions: 

1. Is 30% extra in firmware size worth it? (This is about 30 kB in two dimensions.) Of course, we could add this as an option, selectable via a pre-processor switch, but the code still has to be implemented, and we would have to do something about the documentation: the most-sought feature is the Fourier transform, and if the `complex` can be excluded from the firmware, then there would be two tiers, where the behaviour of the `fft` function depends on the switch. That is highly undesirable. 
2. Could the above-mentioned two feature requests be addressed in a different way, without the introduction of `complex`, as, e.g., in `fft`?
[...]

PR #366, opened 2021-04-20
add complex support

v923z:
This PR adds rudimentary complex support to `ulab`. Detailed discussion can be found in https://github.com/v923z/micropython-ulab/issues/364.

At the moment, complex arrays can be initialised from complex iterables. E.g., 
```python
from ulab import numpy as np
a = np.array([1, 2+2j, 3, 4], dtype=np.complex)
print(a)
```
behaves as expected. In addition, the PR also implements the `real` and imaginary functions for `ndarray`s, i.e., the following works: 

```python
from ulab import numpy as np

a = np.array([1, 2+2j, 3, 4], dtype=np.complex)
b = np.array([1, 2, 3, 4], dtype=np.uint8)

print(a)
print(np.real(a))
print(np.imag(a))
print(np.real(b))
print(np.imag(b))
```


wired8:
Thanks for this.  I've tried running on an ESP32 with MicroPython v1.17 but get kernel panics when doing any operations.
Sample stack trace:
ndarray_make_new_core
ndarray.c:?
ndarray_array_constructor

Any ideas what I may have done wrong?

v923z:
@wired8 This draft is based on an older version of `ulab`, so I am pretty sure that it's not going to fly. Do you need complex support? If so, what would you need it for? This whole issue came up in the context of eigenvalues and eigenvectors, but your requirements could be different, I don't know.

wired8:
@v923z thanks for the quick response.

Specifically I'm trying to port some filters from scipy.

The buttap function uses complex numbers.

```
def buttap(N):
    """Return (z,p,k) for analog prototype of Nth-order Butterworth filter.

    The filter will have an angular (e.g., rad/s) cutoff frequency of 1.

    See Also
    --------
    butter : Filter design function using this prototype

    """
    if abs(int(N)) != N:
        raise ValueError("Filter order must be a nonnegative integer")
    z = numpy.array([])
    m = numpy.arange(-N+1, N, 2)
    # Middle value is 0 to ensure an exactly real pole
    p = -numpy.exp(1j * np.pi * m / (2 * N))
    k = 1
    return z, p, k
  ```

v923z:
The `exp` function could be solved as `cos() + 1j * sin()`, somewhat similar to how the FFT is handled now. But I assume that, after you generate the filter coefficients, you want to pipe your data into `lfilter`, or something similar. Is that correct?
What I am trying to say is that `complex` is most probably only the first stage of what you need.

wired8:
Correct, data is then piped through one of low/high/band or bandstop filter.

v923z:
@wired8 I think I can relatively easily fix the complex code, but if you want to use it, we have to implement the filter functions in question. In order to not invest effort, where it is not needed, could you, please, point out which are the most important functions that you would need. Preferably, in order of importance.

wired8:
https://github.com/v923z/micropython-ulab/issues/447
https://github.com/v923z/micropython-ulab/issues/448
https://github.com/v923z/micropython-ulab/issues/449

Thanks @v923z

v923z:
Closing this now, on-going work is in https://github.com/v923z/micropython-ulab/pull/456

PR #367, opened 2021-04-23
Fix implementation of math constants (e, inf, nan, pi).

t-ikegami:
Representation of float objects depends on MICROPY_OBJ_REPR.  The original implementation of math constants (ulab.numpy.pi and etc.) works for MICROPY_OBJ_REPR = A and B.  It is extended to work with MICROPY_OBJ_REPR = C and D.  Note that CircuitPython basically employs MICROPY_OBJ_REPR_C, so that ulab constants give wrong numbers.

### DIFF
--- a/code/numpy/numpy.c
+++ b/code/numpy/numpy.c
@@ -36,19 +36,47 @@
 
 // math constants
 #if ULAB_NUMPY_HAS_E
+#if MICROPY_OBJ_REPR == MICROPY_OBJ_REPR_C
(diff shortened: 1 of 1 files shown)


v923z:
@t-ikegami Many thanks for fixing the issue, I really appreciate it!

@jepler Jeff, could you, please, acknowledge this PR in the Monday meeting? I will update the `circuitpython` test section in https://github.com/v923z/micropython-ulab/blob/fb44eafb9ced08ab4e23611d2ee414ed7eb15522/build-cp.sh#L48 This is not completely trivial, because properties are different in `micropython`.

PR #370, opened 2021-04-23
Lowercase MP macros

tannewt:
We're updating to newer MP in CircuitPython and will no longer have the upper case versions.


v923z:
@tannewt Scott, this fails with an implicit declaration of the macro: https://github.com/v923z/micropython-ulab/pull/370/checks?check_run_id=2421432960

tannewt:
This is because we haven't updated CircuitPython yet. The pending PR is: https://github.com/adafruit/circuitpython/pull/4649 Once it's merged then we can get this to pass and then update CP back to upstream.

v923z:
> This is because we haven't updated CircuitPython yet. The pending PR is: 

Thanks! I leave this open then. I have subscribed to that thread.

v923z:
@tannewt Thanks a lot, Scott!

Issue #387, opened 2021-05-13
[BUG] ulab array corruption with slices


jepler:
As reported by @kevinjwaters at https://github.com/adafruit/circuitpython/issues/4753, there seems to be a problem with memory corruption when ulab arrays are sliced.

I am in a fresh, up to date clone of ulab master branch (90824d28e5755034d9ba95dd0a6431e4a6d0225e), ran `./build.py`.

I put the following content in `gcproblem.py`:
```
from ulab import numpy as np
import gc
data1 = np.ones(1000)[6:-6]
print(sum(data1)) # should print 988, consistently does
print(data1)      # should print an array full of ones, consistently does
gc.collect()
print(sum(data1)) # should print 988, usually does NOT
print(data1)      # should print an array full of ones, usually does NOT
```
then I run `./micropython/ports/unix/micropython gcproblem.py`:
```
$ ./micropython/ports/unix/micropython  ~/gcproblem.py 
988.0000000000001
array([1.0, 1.0, 1.0, ..., 1.0, 1.0, 1.0], dtype=float64)
14240.0
array([0.0, 0.0, 4.670711089722114e-310, ..., 0.0, 4.670711089722114e-310, 256.0], dtype=float64)
```
The first two lines of output are correct, the second two are wrong.

I notice the following detail about how memoryview works (`py/objarray.c` in the micropython source code):
> memoryview must keep a pointer to the base of the buffer so that the buffer is not GC'd if the original parent object is no longer around (we are assuming that all memoryview'able objects return a pointer which points to the start of a GC chunk).

It may be the case that a form of this trick needs to be followed in ulab.

v923z:
> > memoryview must keep a pointer to the base of the buffer so that the buffer is not GC'd if the original parent object is no longer around (we are assuming that all memoryview'able objects return a pointer which points to the start of a GC chunk).
> 
> It may be the case that a form of this trick needs to be followed in ulab.

So, would it be enough, if, in https://github.com/v923z/micropython-ulab/blob/90824d28e5755034d9ba95dd0a6431e4a6d0225e/code/ndarray.h#L73-L83, we added a new member to the struct, `void *origin`, and then in 
https://github.com/v923z/micropython-ulab/blob/90824d28e5755034d9ba95dd0a6431e4a6d0225e/code/ndarray.c#L767-L785
simply had 
```c
ndarray->origin = source->origin;
```
[...]

PR #388, opened 2021-05-13
fix garbage collection problem

v923z:
This PR should close https://github.com/v923z/micropython-ulab/issues/387


jepler (approved):
it wouldn't hurt if my script was adapted into a test program, but otherwise if it fixes the problem for you I think it's good to go.  I didn't do any testing.

v923z:
> it wouldn't hurt if my script was adapted into a test program

Right. I will see to it. Thanks!

PR #391, opened 2021-05-16
Implementation of cho_solve function in scipy.linalg module

vikas-udupa:
I have added a function to `scipy.linalg` module called `cho_solve`. This function performs the same operation as `scipy.linalg.cho_solve`.

This function solves the equation A x = b for x, given the Cholesky factorization of A, which is referred to as L.
This matrix L and right-hand-side vector b are passed as arguments to the function. It returns the vector x.

Here is a description of the function in scipy docs: https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cho_solve.html

I have also added two tests in `tests/scipy/cho_solve.py`. One of them is to test the function, given the matrix L directly. In the other test, we take a matrix A, find its Cholesky factorization L using `numpy.linalg.cholesky` function, and eventually use L in `scipy.linalg.cho_solve` function to find x.

This is a functionality we wish to have in micropython and I have thus implemented the same. I am open to suggestions/modifications. Thanks.


v923z:
@vikas-udupa This look OK, so I am merging it now. I assume that you didn't want to implement the keyword arguments.

PR #408, opened 2021-06-08
implement ndarray properties

v923z:
This PR implements the `ndarray` properties by means of the `attr` member of the `ulab_ndarray_type`. The original idea was posted in https://github.com/micropython/micropython/pull/5648#issuecomment-831531599

### DIFF
--- a/code/micropython.mk
+++ b/code/micropython.mk
@@ -9,6 +9,7 @@ SRC_USERMOD += $(USERMODULES_DIR)/scipy/special/special.c
 SRC_USERMOD += $(USERMODULES_DIR)/ndarray_operators.c
 SRC_USERMOD += $(USERMODULES_DIR)/ulab_tools.c
 SRC_USERMOD += $(USERMODULES_DIR)/ndarray.c
+SRC_USERMOD += $(USERMODULES_DIR)/ndarray_properties.c
 SRC_USERMOD += $(USERMODULES_DIR)/numpy/approx/approx.c
 SRC_USERMOD += $(USERMODULES_DIR)/numpy/compare/compare.c
 SRC_USERMOD += $(USERMODULES_DIR)/ulab_create.c
--- a/code/ndarray_properties.c
+++ b/code/ndarray_properties.c
@@ -1,0 +1,71 @@
+
+/*
+ * This file is part of the micropython-ulab project,
+ *
+ * https://github.com/v923z/micropython-ulab
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2021 Zoltán Vörös
+ *
+*/
+
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "py/obj.h"
+#include "py/runtime.h"
+
+#include "ulab.h"
+#include "ndarray.h"
+
+#if !CIRCUITPY
+
+// a somewhat hackish implementation of property getters/setters;
+// this functions is hooked into the attr member of ndarray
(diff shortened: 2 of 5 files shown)


v923z:
@jepler Jeff, I have tried to implement the `ndarray` properties for `micropython`, and I thought that I would re-organise the test scripts a bit. `circuitpython` fails with two of them, namely, `numericals`, and `operators`, and in both cases, the root of the problem is how `circuitpython` handles wrap-around values. Here is the traceback

```bash
  File "./tests/numpy/numericals.py", line 18, in <module>
OverflowError: value must fit in 1 byte(s)
```
which is caused by something like this:

```python
a = np.array([range(255-5, 255),range(240-5, 240),range(250-5, 250)], dtype=np.int8)
```
`a` is 
```python
array([[ -6,  -5,  -4,  -3,  -2],
       [-21, -20, -19, -18, -17],
       [-11, -10,  -9,  -8,  -7]], dtype=int8)
```
in both `numpy`, and `micropython-ulab`. 

We didn't happen upon this issue earlier, because this was in a test script that was never run with `circuitpython`. 

Now, the question really is, what we should do about it. Not the test script, but this phenomenon. Can we make this work in `circuitpython`, and if so, is the solution in the core, or we can fix it in `ulab`?

jepler:
We added code that causes `struct.pack('b', 999)` and a similar construct with array to fail, same as standard Python.  However, standard numpy seems to silently ignore upper bits, so re-using that code for these conversions is an incompatibility.
```python
>>> numpy.array([999], dtype=np.int8)
array([-25], dtype=int8)
>>> struct.pack('b', 999)
struct.error: byte format requires -128 <= number <= 127
>>> array.array('b', [999])
OverflowError: signed char is greater than maximum
```

This seems to occur via the in the call to `mp_{small,obj}_int_buffer_overflow_check` called from `mp_binary_set_val_array`. We are not going to be interested in reverting our behavior (it would give up compatibility with standard python array/struct), so options would include adding a flag or a whole separate routine to meet the needs of ulab.

v923z:
> We added code that causes `struct.pack('b', 999)` and a similar construct with array to fail, same as standard Python. However, standard numpy seems to silently ignore upper bits, so re-using that code for these conversions is an incompatibility.

> This seems to occur via the in the call to `mp_{small,obj}_int_buffer_overflow_check` called from `mp_binary_set_val_array`. We are not going to be interested in reverting our behavior (it would give up compatibility with standard python array/struct), 

Should I raise an issue in `micropython`, then?

v923z:
> so options would include adding a flag or a whole separate routine to meet the needs of ulab.

Would something like https://github.com/v923z/micropython-ulab/pull/408/commits/f946abd2ee31bdae9e8f360b10f9a32c6007e51e be OK with `circuitpython`?

jepler:
> > so options would include adding a flag or a whole separate routine to meet the needs of ulab.
> 
> Would something like [f946abd](https://github.com/v923z/micropython-ulab/commit/f946abd2ee31bdae9e8f360b10f9a32c6007e51e) be OK with `circuitpython`?

I think so

v923z:
> > > so options would include adding a flag or a whole separate routine to meet the needs of ulab.
> > 
> > 
> > Would something like [f946abd](https://github.com/v923z/micropython-ulab/commit/f946abd2ee31bdae9e8f360b10f9a32c6007e51e) be OK with `circuitpython`?
> 
> I think so

Thanks! I have removed the clutter. If you are OK with it, you can merge it. I will then create a new tag. Since it is a breaking change on the `micropython` side, I would start a new major version, and set out with 3.0.0.

PR #418, opened 2021-07-06
support circuitpython-style split type objects

jepler:
In CircuitPython, we are working on flash memory savings by "splitting" type objects. These changes support the new split type objects, while also being compatible with implementations that do not have this optimization.

### DIFF
--- a/code/ndarray.h
+++ b/code/ndarray.h
@@ -53,7 +53,7 @@ typedef struct _mp_obj_slice_t {
 #define MP_TYPE_CALL call
 #define mp_type_call(t) t->call
 #define MP_TYPE_FLAG_EXTENDED (0)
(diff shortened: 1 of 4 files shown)


jepler:
There's a difference in the level of support for static initialization of flexible array members in gcc vs clang.
```c
typedef struct flex {
   int i;
   struct {
       int a, b, c;
   } j[];
} flex_t;

flex_t f1 = {.i = 3};
flex_t f2 = {.i = 4, .j[0].a = 1, };
flex_t f3 = {.i = 4, .j = {{.a = 1,}}, };
```

Styles f1 and f3 are supported in gcc and clang; style f2 is supported in gcc but not in clang.  Therefore, we must select style f3 when extended fields are being initialized.

v923z:
@jepler Thanks!

Issue #421, opened 2021-07-08
[FEATURE REQUEST]add flat iterator for higher-dimensional tensors


v923z:
As discussed in https://github.com/v923z/micropython-ulab/issues/420, it would be useful to have an iterator that can traverse a higher-dimensional tensor without copying the underlying data. 

`numpy`: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flat.html
https://numpy.org/doc/stable/reference/generated/numpy.flatiter.html

v923z:
https://github.com/v923z/micropython-ulab/pull/422 closes the issue.

Issue #420, opened 2021-07-08
Q: How to iterate over all elements of a 2D matrix in unspecified order


nsajko:
I have no experience with ulab (nor numpy, for that matter), and I couldn't find a discussion forum dedicated to ulab, so i hope it's OK to ask the question here.

So I have an ndarray like this:

```
a = np.array([[1, 2],[7, 5]], dtype = np.float)
```

and I want to iterate over all of its elements, so over the set `{1, 2, 5, 7}`.

I think that in numpy one would use `a.flat` for that purpose, but that doesn't seem to be available in ulab.

What is the intended way to solve this problem?

v923z:
> I have no experience with ulab (nor numpy, for that matter), and I couldn't find a discussion forum dedicated to ulab, so i hope it's OK to ask the question here.

It's OK. 

> 
> So I have an ndarray like this:
> 
> ```
> a = np.array([[1, 2],[7, 5]], dtype = np.float)
> ```
> 
> and I want to iterate over all of its elements, so over the set `{1, 2, 5, 7}`.
> 
> I think that in numpy one would use `a.flat` for that purpose, but that doesn't seem to be available in ulab.

You have at least two options. One is the [.flatten method](https://micropython-ulab.readthedocs.io/en/latest/ulab-ndarray.html?highlight=flatten#flatten), the other is [.reshape](https://micropython-ulab.readthedocs.io/en/latest/ulab-ndarray.html?highlight=flatten#reshape).

nsajko:
It's very possible I'm wrong, but I was under the impression that `.flat` avoids allocations that would happen with `.flatten`?

Is there a way to iterate over all elements of a matrix in a single loop, but without copying the backing storage of the matrix?

v923z:
> It's very possible I'm wrong, but I was under the impression that `.flat` avoids allocations that would happen with `.flatten`?
> 
> Is there a way to iterate over all elements of a matrix in a single loop, but without copying the backing storage of the matrix?

`.reshape` does not copy anything, if the matrix is dense. It will return a view of the underlying data. 

But, if you really want to have a flat iterator, it is definitely doable, and it's not even that hard. If you are not happy with `.reshape`, I could add the flat iterator in a week or so, perhaps, even sooner.

nsajko:
Don't do it just for me :)

Thank you, I will look into reshape, then.

v923z:
> Don't do it just for me :)

[...]

PR #422, opened 2021-07-13
Flatiter

v923z:
This PR implements the feature request https://github.com/v923z/micropython-ulab/issues/421, https://github.com/v923z/micropython-ulab/issues/420

### DIFF
--- a/code/micropython.mk
+++ b/code/micropython.mk
@@ -9,6 +9,7 @@ SRC_USERMOD += $(USERMODULES_DIR)/scipy/special/special.c
 SRC_USERMOD += $(USERMODULES_DIR)/ndarray_operators.c
 SRC_USERMOD += $(USERMODULES_DIR)/ulab_tools.c
 SRC_USERMOD += $(USERMODULES_DIR)/ndarray.c
+SRC_USERMOD += $(USERMODULES_DIR)/numpy/ndarray/ndarray_iter.c
 SRC_USERMOD += $(USERMODULES_DIR)/ndarray_properties.c
 SRC_USERMOD += $(USERMODULES_DIR)/numpy/approx/approx.c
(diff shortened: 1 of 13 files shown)


v923z:
@jepler Jeff, this fails in `circuitpython`, and I don't know, where exactly, and why. But, perhaps, the question is, whether you want to have this in the first place: https://github.com/v923z/micropython-ulab/blob/flatiter/docs/manual/source/ulab-ndarray.rst#flat .

We can also think about using attributes, as in `micropython`. If you still want to use properties, then we might have to insert the setter in some of the cases. E.g., in `micropython`, you can now reshape an array by writing into the `.shape` attribute: https://github.com/v923z/micropython-ulab/blob/e881ccb30c0c52610188d3c30779010dc3a6eb0f/code/ndarray_properties.c#L83-L89, https://micropython-ulab.readthedocs.io/en/latest/ulab-ndarray.html#shape

jepler:
Quick answer about how properties are implemented: While we have a specific way of doing it in CircuitPython core, we don't mind if ulab implements it in another way as long as the same code works.

Longer answer about `getiter`: This has to do with https://github.com/adafruit/circuitpython/pull/4903 -- I added a commit to this branch which should resolve it.

v923z:
> Quick answer about how properties are implemented: While we have a specific way of doing it in CircuitPython core, we don't mind if ulab implements it in another way as long as the same code works.

OK. I will merge it now as it is, and clean up the code later. I can also try to see if we can save some flash by employing properties vs. attributes. 

> 
> Longer answer about `getiter`: This has to do with [adafruit/circuitpython#4903](https://github.com/adafruit/circuitpython/pull/4903) -- I added a commit to this branch which should resolve it.

Many thanks!

Issue #419, opened 2021-07-08
[BUG] CircuitPython documentation appears in wrong location (ulab.X vs ulab.numpy.X)


jepler:
**Describe the bug**
In CircuitPython main branch, which is based off 161a7288482200c884623ed3cb59dd91f53be91a (3.0.0), many items appear in the wrong location in the documentation.  For instance, the documentation shows that the `array` constructor is at `ulab.array` rather than `ulab.numpy.array`.

**To Reproduce**
View the documentation on [readthedocs](https://circuitpython.readthedocs.io/en/latest/shared-bindings/ulab/index.html#ulab.array) or use `./build-cp.sh` locally and look at the generated docs

**Expected behavior**
Items are organized in the documentation as they are in the implementation

v923z:
Is there an easy way to generate a comparison between what it is, and what it should be?

jepler:
No easy way to generate lists like that comes to mind.

I just know that almost nothing is directly in `ulab` anymore, and almost everything is down in submodules like `ulab.numpy`.

v923z:
I don't want to combine everything that doesn't belong to sub-modules into a single file, but would something like this work?

```
ulab
|----numpy
|    |----poly
|    |     |----poly.c
|    |     |----poly.h
|    |----numerical
|    |     |----numerical.c
|    |     |----numerical.h
|    |----transform
|    |     |----transform.c
|    |     |----transform.h
|    |----approx
|    |     |----approx.c
|    |     |----approx.h
|    |----modules
|         |----fft
|         |     |----fft.c
|         |     |----fft.h
|         |----linalg
|               |----linalg.c
|               |----linalg.h
|
|----scipy
|    |----modules
|         |----signal
|         |----linalg
|         |----optimize
|
|----utils
|
|----user
```
An alternative could be 

```
ulab
|----numpy
|    |----core
|    |     |----poly.c
|    |     |----poly.h
|    |     |----transform.c
|    |     |----transform.h
|    |     |----approx.c
|    |     |----approx.h
|    |----modules
|         |----fft
|         |     |----fft.c
|         |     |----fft.h
|         |----linalg
|               |----linalg.c
|               |----linalg.h
|
|----scipy
|    |----modules
|         |----signal
|         |----linalg
|         |----optimize
|
|----utils
|
|----user
```
[...]

PR #423, opened 2021-07-14
[WIP] re-arrange files, so that circuitpython docs can be generated

v923z:
Should address https://github.com/v923z/micropython-ulab/issues/419

PR #426, opened 2021-07-21
Circuitpython doc move

jepler:
This builds on #423 and also works together with https://github.com/adafruit/circuitpython/pull/5028 to complete the doc move. (If we take this PR, then #423 is moot)

However, this build will come up in red for the circuitpython builds because without https://github.com/adafruit/circuitpython/pull/5028 (which renames some use of ulab.ndarray to ulab.numpy.ndarray within our own pyi/docs) it is still not OK.

I didn't find a way for both things to work together, so the possibility I see is to (approximately in order):

- [ ] verify that https://github.com/adafruit/circuitpython/pull/5028 is green & approved
- [x] verify that this PR is green for micropython & approved
- [ ] merge this PR
- [ ] release ulab
- [ ] Update https://github.com/adafruit/circuitpython/pull/5028 with the released ulab
- [ ] merge https://github.com/adafruit/circuitpython/pull/5028

then we can all be green again.

Closes: #419


v923z:
@jepler  OK, I will look at it later today. Thanks for following up on this!

v923z:
@jepler Jeff, thanks again for taking care of this! I don't see any problems here, so we can merge the PR. (I don't know if your list is chronological, i.e., whether the approval of the first point must precede that of the second). I can prepare a release, no problem. I believe, there are two new functions/methods since the last release.

jepler:
It's semi-chronological, so green + approved on the circuitpython PR is what I'd wait for next.

jepler:
I think I found a way that gives forward compatibility, so if this turns out green I'll merge it.

v923z:
> I think I found a way that gives forward compatibility, so if this turns out green I'll merge it.

Fine with me.

jepler:
@v923z yay -- ping me when you have had time to make a release and I'll start the ball rolling in circuitpython to pull that version in. Thanks for doing the initial work on this doc stuff as well!

v923z:
@jepler I can create a tag before 1pm, your time.

PR #429, opened 2021-07-22
Build all dimensions

jepler:
... actually, build 2, 3, and 4 dimensions.  I'll file a separate issue showing why I didn't enable building 1D.

This:
 * changes the build scripts so that they take a positional argument, optional, defaulting to 2: `./build.sh 3`
 * adds some dimension-linked tests;  Say you're doing a 3D build, it will run the 1d, 2d, and 3d tests in addition to the "non-dimensoined" tests.

I suspect some existing tests will need to be moved into 2d; we can also introduce a 2-level structure if it makes sense, e.g., `2d/scipy`, `3d/numpy`, or whatever is most useful.  No existing tests had to be moved to 2d in part because of not enabling a 1d build in the CI yet.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
             os:
                 - ubuntu-16.04
                 - macos-10.14
+            dims: [2, 3, 4]
     runs-on: ${{ matrix.os }}
     steps:
     - name: Dump GitHub context
@@ -46,7 +47,7 @@ jobs:
         path: micropython
 
     - name: Run build.sh
-      run: ./build.sh
+      run: ./build.sh ${{ matrix.dims }}
 
   circuitpython:
     strategy:
@@ -54,6 +55,7 @@ jobs:
             os:
                 - ubuntu-20.04
                 - macos-10.14
+            dims: [2, 3, 4]
     runs-on: ${{ matrix.os }}
     steps:
     - name: Dump GitHub context
@@ -85,4 +87,4 @@ jobs:
         python3 -mpip install -r requirements_cp_dev.txt
 
     - name: Run build-cp.sh
-      run: ./build-cp.sh
+      run: ./build-cp.sh ${{ matrix.dims }}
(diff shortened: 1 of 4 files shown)


v923z:
@jepler  Thanks, this is really a huge improvement. 

> ... actually, build 2, 3, and 4 dimensions. I'll file a separate issue showing why I didn't enable building 1D.
> 
> This:
> 
>     * changes the build scripts so that they take a positional argument, optional, defaulting to 2: `./build.sh 3`
> 
>     * adds some dimension-linked tests;  Say you're doing a 3D build, it will run the 1d, 2d, and 3d tests in addition to the "non-dimensoined" tests.
> 
> 
> I suspect some existing tests will need to be moved into 2d; we can also introduce a 2-level structure if it makes sense, e.g., `2d/scipy`, `3d/numpy`, or whatever is most useful. 

I would like to have a single source of truth for the scripts: the dimensionality can be found out

```python
import ulab
version = ulab.__version__
version_dims = version.split('-')[1]
```
so we can conditionally execute chunks of the code. However, the expected files do need to have 4 versions, I believe. Do you think we could pipe the results into `some_test.exp.1D`, `some_test.exp.2D`, `some_test.exp.3D, `some_test.exp.4D`? However, if you want to, we could just have `numpy/1D`, ``numpy/2D` etc., I am not against that. These are generated files, it doesn't really matter. Your approach might actually be tidier. My comment on the single source applies only to the script itself.

jepler:
My assumption was this: A test _requires_ at least N dimensions to run, but if it can be run it gives the same results even if M>N dimensions are supported.  Hence the idea of organizing them by the (minimum) dimensionality they require.

Another way is to encode the skip condition into the test. Any test can print "SKIP" and exit with success to be skipped.  You could do this based on explicit information from the version string (or from a new property such as ulab.max_dim), or by trying to construct an array of the desired dimensionality and handling the exception:

```python
from ulab import numpy as np
try:
    np.ones((1,1,1))
except TypeError:
    print("SKIP")
    raise SystemExit

# test requiring at least 3 dims goes here
```

Let me know how you'd like to organize the tests and I'll implement what goes in the build.sh files.

jepler:
Closing in favor of #432

Issue #430, opened 2021-07-22
[BUG] Building for "1D" doesn't work


jepler:
**Describe the bug**
With ulab 3.3.1 + #429, `./build.sh 1` (to build only with support for 1D arrays) fails at the compile step

**To Reproduce**
On a host computer, run `./build.sh 1` (or `./build-cp.sh 1`).

**Expected behavior**
Builds and passes tests

**Additional context**
Here are the errors I receive locally:
```
jepler@babs:~/src/ulab$ ./build.sh 1
make: Entering directory '/home/jepler/src/ulab/micropython/mpy-cross'
Use make V=1 or set BUILD_VERBOSE in your environment to increase build verbosity.
make: Leaving directory '/home/jepler/src/ulab/micropython/mpy-cross'
make: Entering directory '/home/jepler/src/ulab/micropython/ports/unix'
Use make V=1 or set BUILD_VERBOSE in your environment to increase build verbosity.
make: Nothing to be done for 'axtls'.
make: Leaving directory '/home/jepler/src/ulab/micropython/ports/unix'
make: Entering directory '/home/jepler/src/ulab/micropython/ports/unix'
Use make V=1 or set BUILD_VERBOSE in your environment to increase build verbosity.
Including User C Module from /home/jepler/src/ulab/code
GEN build-1/genhdr/qstr.i.last
/home/jepler/src/ulab/code/numpy/numerical.c:324:100: error: macro "RUN_MEAN_STD1" passed 7 arguments, but takes just 6
  324 | p_float_t, array, farray, _shape_strides, MICROPY_FLOAT_CONST(0.0), 0);
      |                                                                      ^

In file included from /home/jepler/src/ulab/code/numpy/numerical.c:25:
/home/jepler/src/ulab/code/numpy/numerical.h:89: note: macro "RUN_MEAN_STD1" defined here
   89 | #define RUN_MEAN_STD1(type, array, rarray, ss, div, isStd)\
      | 
/home/jepler/src/ulab/code/numpy/numerical.c:342:80: error: macro "RUN_MEAN_STD1" passed 7 arguments, but takes just 6
  342 |       RUN_MEAN_STD(uint8_t, array, farray, _shape_strides, div, isStd);
      |                                                                      ^

In file included from /home/jepler/src/ulab/code/numpy/numerical.c:25:
/home/jepler/src/ulab/code/numpy/numerical.h:89: note: macro "RUN_MEAN_STD1" defined here
[...]

PR #431, opened 2021-07-23
fix compilation for 1D

v923z:
This PR re-introduces the option of compiling with ULAB_MAX_DIMS = 1, and should close https://github.com/v923z/micropython-ulab/issues/430 .


jepler (commented):
Nice! I also pulled this into #432 but there's no reason it can't be merged right away.

PR #432, opened 2021-07-23
Build & test for 1, 2, 3, and 4 dims

jepler:
This pulls in #431 and #429 both.

Tests are organized as `tests/1d/numpy`, `tests/2d/scipy`, etc, so that the _minimum_ dimensionality and the _general area_ of the test are both captured in the test's path.  I moved some tests that I could into `1d` from `2d`.


v923z:
@jepler Thanks for taking the initiative on this! If you are satisfied, you should just go ahead, and merge this. We can extend the tests later, if there are cases that would have to be checked for 3D, and 4D.

PR #444, opened 2021-11-05
Fix compare between (uint16|int16) vs float.

t-ikegami:
This PR fixes the following failure in ulab.numpy.maximum (and etc., observed in CircuitPython 7.0).

```
>>> np.maximum(np.ndarray([1000], dtype = np.uint16), 10.0)
array([232.0], dtype=float32)
>>> np.maximum(np.ndarray([-1000], dtype = np.int16), 10.0)
array([64536.0], dtype=float32)
```


v923z:
@t-ikegami Many thanks for the PR! I think this is OK, but before merging, I would like to request two small changes. First, set the version to 3.3.5 in https://github.com/v923z/micropython-ulab/blob/a2455ca7cedf92be7b32ef0f710c694f9ddef65d/code/ulab.c#L36, and add an entry in https://github.com/v923z/micropython-ulab/blob/master/docs/ulab-change-log.md.

v923z:
@t-ikegami Many thanks for contributing!

PR #446, opened 2021-11-07
Fix Makefile misspelling

JorgeGMarques:
esp32-cmake.sh was building the firmware without the ulab module because it was not patching the Make**f**ile, but trying to patch the nonexistent Make**F**ile.

### DIFF
--- a/build/esp32-cmake.sh
+++ b/build/esp32-cmake.sh
@@ -26,10 +26,10 @@ cd $BUILD_DIR/micropython/ports/esp32
 make submodules
 
 echo "--- PATCH MAKEFILE ---"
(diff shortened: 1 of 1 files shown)


mdaeron:
Oops. Sorry for the typo, I tested that on macOS which silently ignores the case difference.

v923z:
@JorgeGMarques Jorge, is this patch ready for merge?

Thanks for contributing!

JorgeGMarques:
Yes, I tested the compiled firmware on my ESP32, it included the ulab module on it after the patch :smile:

v923z:
@JorgeGMarques Fantastic, thanks for the feedback!

Issue #443, opened 2021-11-04
[FEATURE REQUEST] Enable versions in ReadTheDocs documentation


RobertoRoos:
**Describe the solution you'd like**

In short: can different older tagged versions be enabled in the RTD project: https://readthedocs.org/projects/micropython-ulab

In the admin panel you can enable which branches and tags you want built and available. I would like to access the docs of some older versions, because we have a old fork of this repo running somewhere.

I understand this is quite niche, but considering how little effort it takes I hope this can be considered!

RobertoRoos:
In the meantime I made a fresh fork with my own RTD project, this is what it would then look like: https://micropython-ulab-robert.readthedocs.io/en/0.36.0

v923z:
@RobertoRoos Thanks for raising the issue. I also think that this feature would be useful. I'll fix the RTD side later today.

v923z:
@RobertoRoos I have activated the older versions on RTD, but all prior to 2.1.2 fail, because at that time we stipulated that the requirements file should be used, and that does not exist for the old tags. I don't want to modify those versions on github. Do you know, whether we could fix this in RTD itself? I would be OK with a one-time setting of version-specific build options, if that is possible.

RobertoRoos:
Hm, in my fork it seems only the v1.* versions fail, v0.* do work: https://readthedocs.org/projects/micropython-ulab-robert/builds/ (but I haven't enabled many). I kept all the default settings in RTD (so no requirements.txt path).

This is why the [.readthedocs.yaml](https://docs.readthedocs.io/en/stable/config-file/index.html) file is convenient, over the online settings.

We could commit the current RTD settings to the YAML file for current and future settings, and tweak the online settings to work with legacy versions?
Actually, you could probably reset the online settings too after those builds are completed.

But I could understand this might be more effort than it's worth. Thanks for versions that did built, that might be enough for me!

v923z:
> Hm, in my fork it seems only the v1.* versions fail, v0.* do work: https://readthedocs.org/projects/micropython-ulab-robert/builds/ (but I haven't enabled many). I kept all the default settings in RTD (so no requirements.txt path).

We had to add the requirements at one point, because the defaults didn't load the autoapi. 

> 
> This is why the [.readthedocs.yaml](https://docs.readthedocs.io/en/stable/config-file/index.html) file is convenient, over the online settings.

[...]

PR #450, opened 2021-11-19
Created .readthedocs.yaml config file for latest versions

RobertoRoos:
This should replace all settings from readthedocs.org. This is the recommended approach, as it's controlled by versions then.

This version is built at: https://micropython-ulab-robert.readthedocs.io/en/feature-rtd_config/

Relates to #443


v923z:
Thanks!

PR #453, opened 2021-11-19
Adding branch release/v1.7 to suit version 1.7.9

RobertoRoos:
New branch needs to be created and PR needs to be retargeted.

Example build is here: https://micropython-ulab-robert.readthedocs.io/en/1.7.9/

Relates to #443

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,25 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.9"
+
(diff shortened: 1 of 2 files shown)


RobertoRoos:
I updated with the latest `conf.py` and now it builds. I have to admit I haven't looked at the built output in detail, I just assume the output fails or is correct.

RobertoRoos:
Nonono, this isn't supposed to go into the `legacy` branch. All these versions should have their own branch, but I don't have the power to create branches here.

Can you force push this merge out? And then retarted this PR?

v923z:
@RobertoRoos I think 1.7.9 is OK here, because 1.7.8 was at the end of the legacy branch. So, in that case, we can continue there in a linear fashion. 
On the other hand, 0.36, and 0.54.4 should have their own branches, which I created.

RobertoRoos:
@v923z yeah I see now. Thanks for merging!

Yeah, I'm having my doubts about those versions now 😅
Funny thing is we now have the ticket to also update our plan version...

PR #456, opened 2021-11-23
Complex2

v923z:
This is the re-based version of https://github.com/v923z/micropython-ulab/pull/366, which will be closed now.


v923z:
@wired8 Could you, please, check, whether you still have the error that you mentioned in https://github.com/v923z/micropython-ulab/pull/366#issuecomment-965685473 ?

wired8:
> @wired8 Could you, please, check, whether you still have the error that you mentioned in [#366 (comment)](https://github.com/v923z/micropython-ulab/pull/366#issuecomment-965685473) ?

Looks good, no more kernal panic.  

Doing some testing with the following code:

```
>>> N = 5
>>> z = numpy.array([])
>>> m = numpy.arange(-N+1, N, 2)
>>> p = -numpy.exp(1j * np.pi * m / (2 * N))
TypeError: 'complex' object isn't iterable
```

Can we add support for iterables over complex types?

v923z:
@wired8 

> Doing some testing with the following code:
> 
> ```
> >>> N = 5
> >>> z = numpy.array([])
> >>> m = numpy.arange(-N+1, N, 2)
> >>> p = -numpy.exp(1j * np.pi * m / (2 * N))
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> TypeError: 'complex' object isn't iterable
> ```
> 
> Can we add support for iterables over complex types?

The problem is that the exponential function is not defined for complexes at the moment. I could implement that, but the question really is, what would be useful in this regard. There are some 30 universal functions, and I am not sure, whether adding everything makes sense. The exponential function should definitely be included. What else? Here is a rudimentary list.

- [x] unary
- [ ] binary
- [ ] exp
- [x] imag
- [x] real
- [ ] FFT in a user-configurable way, so that complete `numpy`-compatibility can be achieved.

wired8:
The list you provided is a great start.  Implementing binary, exp would give us partial butterworth filter support.

wired8:
>>> np.exp(-3.141592653589793j)
(-1+8.742278e-08j)

>>> np.exp([-3.141592653589793j])
TypeError: can't convert complex to float

v923z:
> > > > np.exp(-3.141592653589793j)
> > > > (-1+8.742278e-08j)
> 
> > > > np.exp([-3.141592653589793j])
> > > > Traceback (most recent call last):
> > > > File "", line 1, in 
> > > > TypeError: can't convert complex to float

I think this is OK that way, at least, intentional. If you need the exponential of a complex list, you can cast it first as 

```
a = np.array([3.14j], dtype=np.complex)
np.exp(a)
```

The reason for this is that, when applying a function to an object, first you have to figure out, 
whether the output is going to be complex, or floating point. But you can learn that only by inspecting 
each element, so you can do that in a double pass only. I am not sure it's worth the trouble. 

Floats are simpler in this regard, because applying a mathematical function to a number always 
produces a float, so you can automatically assume that the output is of float dtype.

wired8:
```python
>>> x = (-0.00425085+0.01026246j)
>>> wo = 0.03141673
>>> x + np.sqrt(x**2 - wo**2)   
TypeError: 'complex' object isn't iterable

**Comment by v923z** [comment]
> ```python
> >>> x = (-0.00425085+0.01026246j)
> >>> wo = 0.03141673
> >>> x + np.sqrt(x**2 - wo**2)   
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> TypeError: 'complex' object isn't iterable
> ```

The `sqrt` function is not implemented for complex numbers. Only the `exp` function is.

**Comment by wired8** [comment]
Any chance we can get sqrt support?

**Comment by v923z** [comment]
@wired8 
> Any chance we can get sqrt support?

I am definitely not against it, but first you should suggest a solution to the following problem. 

The exponential of a real array is always real, while the exponential of a complex array is complex. Now, the square root of a complex array is complex, but the square root of a real array is not necessarily real. 

Say, you have 1000 floats, the first 999 are positive, while the last one is negative. That would require you to return a complex array, when you pass the array to `sqrt`. However, in order to know this, you have to inspect 1000 numbers first. Is that acceptable?

**Comment by wired8** [comment]
I don't see any other choice, unless you make the input explicit and throw exception if user provides incorrect type.

e.g. np.sqrt([0.2, 0.5, 0.4j], dtype=np.complex)

**Comment by v923z** [comment]
> I don't see any other choice, unless you make the input explicit and throw exception if user provides incorrect type.
> 
> ```
> e.g. np.sqrt([0.2, 0.5, 0.4j], dtype=np.complex)
> ```

OK, but how do you know that the `dtype` is incorrect without inspecting all elements, one by one?

**Comment by wired8** [comment]
You don't check, just try/catch throw if wrong.

**Comment by v923z** [comment]
> You don't check, just try/catch throw if wrong.

OK, then what would you specify as the default `dtype`? `float`, as now?

**Comment by wired8** [comment]
Float sounds like the right choice.

**Comment by v923z** [comment]
What makes me a bit uneasy here is the fact that the behaviour of the `exp` and `sqrt` functions would be different. For `exp`, you don't have to specify the `dtype`, while for `sqrt` you do.

**Comment by wired8** [comment]
The only other option is to loop through the array looking for a complex type, exit on find.   This is cleaner, but performance penalty.

What about a hybrid approach....  If the use specifies type then you can skip the check?

**Comment by v923z** [comment]
> What about a hybrid approach.... If the user specifies type then you can skip the check?

Could be done. I would allow only `float`, and `complex` as the output type. I believe, `numpy` doesn't allow integers, either. Give me one or two days to iron out the details. I am definitely working on it.

**Comment by wired8** [comment]
Awesome @v923z appreciate all the hard work.  Thank you.

**Comment by wired8** [comment]
Small bug, the real part of a complex array is not multiplied:

```python
>>> print ( -0.3826834+0.9238795j * 0.5 )
(-0.3826834+0.4619398j)

Should return -0.1913417 +0.46193975j

**Comment by v923z** [comment]
> Small bug, the real part of a complex number is not multiplied:
> 
> ```python
> >>> print ( -0.3826834+0.9238795j * 0.5 )
> (-0.3826834+0.4619398j)
> 
> Should return -0.1913417 +0.46193975j
> ```
> 
> NOTE: Same issue for division operation

This has nothing to do with `ulab`: you are using the `micropython` multiplication operator. `-0.3826834+0.9238795j` wouldn't be an array, anyway. 

I am not done yet with the implementation of the ndarray's binary operators. See https://github.com/v923z/micropython-ulab/pull/456#issuecomment-983980418 It's not quite trivial, so I need a couple of days to do it properly.

**Comment by wired8** [comment]
Thanks @v923z

**Comment by v923z** [comment]
@jepler Jeff, this PR fails with `circuitpython` 2D. What is really weird is that the error comes from the documentation: https://github.com/v923z/micropython-ulab/runs/4653072586?check_suite_focus=true. How can that be dimension-dependent? I have re-run the jobs in the hope that this was a singular github glitch, but that is actually not the case. Any thoughts? 

Otherwise, this is almost done, and if the checks pass, I would like to roll out complex support. I understand that `circuitpython` is not interested in that, and if you don't need it, then nothing (not even the firmware size) changes.

**Comment by v923z** [comment]
@wired8 Phil, do you think you could help with testing? If so, what is your time frame? If you could lend a helping hand here, we could release complex support before the end of the year.

**Comment by wired8** [comment]
> @wired8 Phil, do you think you could help with testing? If so, what is your time frame? If you could lend a helping hand here, we could release complex support before the end of the year.

No problem.  I'll do some on device testing later today and submit results.

**Comment by v923z** [comment]
> > @wired8 Phil, do you think you could help with testing? If so, what is your time frame? If you could lend a helping hand here, we could release complex support before the end of the year.
> 
> No problem. I'll do some on device testing later today and submit results.

Oh. Actually I meant proper tests, like these: https://github.com/v923z/micropython-ulab/tree/master/tests

**Comment by v923z** [comment]
@jepler Sorry for the noise, https://github.com/v923z/micropython-ulab/commit/71b3ba15af48f2c9332d4f257a35157d8e9de953 fixes the issue. Since this is only two lines, I simply copied the code, and didn't go through proper rebasing. I hope you won't mind.

**Comment by jepler** [comment]
thanks, I'm glad you were able to resolve it.

**Comment by wired8** [comment]
Hey @v923z can you grant me repo access so I can submit a PR for tests?

**Comment by wired8** [comment]
Hi @v923z, made some good progress on my end implementing a filter.  At a point where I need the to get the real and complex parts of an array.  Any progress on these functions?

**Comment by wired8** [comment]
np.ones is broken.

```python
>>>print(np.ones(4, dtype=np.uint8))
array([255, 255, 255, 255], dtype=uint8)

>>>print(np.ones(4, dtype=np.complex))
array([-0.0+-0.0j, -0.0+-0.0j, -0.0+-0.0j, -0.0+-0.0j], dtype=complex)

**Comment by v923z** [comment]
> np.ones is broken.
> 
> ```python
> >>>print(np.ones(4, dtype=np.uint8))
> array([255, 255, 255, 255], dtype=uint8)
> 
> >>>print(np.ones(4, dtype=np.complex))
> array([-0.0+-0.0j, -0.0+-0.0j, -0.0+-0.0j, -0.0+-0.0j], dtype=complex)
> ```

None of the functions generating an array (`ones`, `zeros`, `linspace`, `arange` etc.) is implemented for complexes. But we should definitely raise an exception at least. 

However, you can do this:

```
ones(4) + 0j
```

**Comment by v923z** [comment]
> np.ones is broken.
> 
> ```python
> >>>print(np.ones(4, dtype=np.uint8))
> array([255, 255, 255, 255], dtype=uint8)
> 
> >>>print(np.ones(4, dtype=np.complex))
> array([-0.0+-0.0j, -0.0+-0.0j, -0.0+-0.0j, -0.0+-0.0j], dtype=complex)
> ```

@wired8  I have fixed up `ones`, `zeros`, and `full` for the complex case. Anything else?

**Comment by wired8** [comment]
Yes... I'm trying to implement signal.zpk2tf  - https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.zpk2tf.html which calls the following:

numpy.poly - https://github.com/numpy/numpy/blob/v1.21.0/numpy/lib/polynomial.py#L44-L165

I tried writing an implementation of this from its source, but there are too many methods not implemented for complex.

numpy.convolve
numpy.conjugate
numpy.compress
numpy.sort_complex

**Comment by v923z** [comment]
> numpy.convolve

We can definitely extend `convolve` for the complex case. 

> numpy.conjugate

This is trivial, although, you can already do this:

```
a = np.array(..., dtype=np.complex)
b = real(a) - imag(a) * 1j
```

> numpy.compress

Is it this one https://numpy.org/doc/stable/reference/generated/numpy.compress.html?

> numpy.sort_complex

This could be added. As far as I see, this works with 1D arrays only, so the implementation is almost trivial.

**Comment by wired8** [comment]
> > numpy.convolve
> 
> We can definitely extend `convolve` for the complex case.
> 
> > numpy.conjugate
> 
> This is trivial, although, you can already do this:
> 
> ```
> a = np.array(..., dtype=np.complex)
> b = real(a) - imag(a) * 1j
> ```
> 
> > numpy.compress
> 
> Is it this one https://numpy.org/doc/stable/reference/generated/numpy.compress.html?

Yes
> 
> > numpy.sort_complex
> 
> This could be added. As far as I see, this works with 1D arrays only, so the implementation is almost trivial.

This is the guts of the zpk2tf method, so if we can implement these methods we should be able to support this function which is the last part of the iirfilter.

```python
 if issubclass(b.dtype.type, numpy.complexfloating):
        # if complex roots are all complex conjugates, the roots are real.
        roots = numpy.asarray(z, complex)
        pos_roots = numpy.compress(roots.imag > 0, roots)
        neg_roots = numpy.conjugate(numpy.compress(roots.imag < 0, roots))
        if len(pos_roots) == len(neg_roots):
            if numpy.all(numpy.sort_complex(neg_roots) ==
                         numpy.sort_complex(pos_roots)):
                b = b.real.copy()

**Comment by v923z** [comment]
> ```python
>  if issubclass(b.dtype.type, numpy.complexfloating):
>         # if complex roots are all complex conjugates, the roots are real.
>         roots = numpy.asarray(z, complex)
>         pos_roots = numpy.compress(roots.imag > 0, roots)
>         neg_roots = numpy.conjugate(numpy.compress(roots.imag < 0, roots))
>         if len(pos_roots) == len(neg_roots):
>             if numpy.all(numpy.sort_complex(neg_roots) ==
>                          numpy.sort_complex(pos_roots)):
>                 b = b.real.copy()
> ```

Should `real/imag` be properties of the `ndarray`? Otherwise, you have to do
```
b = real(b)
```
That also copies the content automatically.

wired8:
> > ```python
> >  if issubclass(b.dtype.type, numpy.complexfloating):
> >         # if complex roots are all complex conjugates, the roots are real.
> >         roots = numpy.asarray(z, complex)
> >         pos_roots = numpy.compress(roots.imag > 0, roots)
> >         neg_roots = numpy.conjugate(numpy.compress(roots.imag < 0, roots))
> >         if len(pos_roots) == len(neg_roots):
> >             if numpy.all(numpy.sort_complex(neg_roots) ==
> >                          numpy.sort_complex(pos_roots)):
> >                 b = b.real.copy()
> > ```
> 
> Should `real/imag` be properties of the `ndarray`? Otherwise, you have to do
> 
> ```
> b = real(b)
> ```
> 
> That also copies the content automatically.

Keeps it to spec with numpy, so why not!

v923z:
> > Should `real/imag` be properties of the `ndarray`? Otherwise, you have to do
> > ```
> > b = real(b)
> > ```
> > That also copies the content automatically.
> 
> Keeps it to spec with numpy, so why not!

Well, because if you want to have `numpy`, then use `numpy`! Implementing extra features (i.e., those whose functionality can be gotten by other means) adds to the firmware size, eats RAM, and consumes development resources. By the way, I think you don't need to copy in `b.real.copy()`, because `b.real` should return a new copy, and not a view. At least, if I read the specs correctly: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.real.html. 

But to address your comment, `b.real` seems to be a simply alias for `real(b)`, so we could add it with little overhead. `b.imag`, too.

wired8:
@v923z  awesome work, I am now able to get the correct output for a butter filter with one small caveat that the all function doesn't support complex yet.  I do have a work around for now.   https://micropython-ulab.readthedocs.io/en/latest/numpy-functions.html#all

v923z:
> @v923z awesome work, I am now able to get the correct output for a butter filter with one small caveat that the all function doesn't support complex yet. 

@wired8 There, you have it: https://github.com/v923z/micropython-ulab/pull/456/commits/d1b3d40b475a01f59eaaecea5d538d3e78cdf9c6

wired8:
Possible bug with np.all

```python
p = np.array([1j,2j,3j,4j], dtype=np.complex)
q = np.array([1j,2j,3j,4j], dtype=np.complex)
print(np.all(p==q))

>>> False

wired8:
Possible bug with sorting and conjugate

```python
roots = np.array([0.98-0.01j,0.99+0.01j,0.98+0.01j,0.99-0.01j], dtype=np.complex)
p = np.sort_complex(roots)
q = np.sort_complex(np.conjugate(roots))
print(p)
print(q)

>>> array([0.9800000000000001-0.01j, 0.9800000000000001+0.01j, 0.99-0.01j, 0.99+0.01j], dtype=complex)
>>> array([0.9800000000000001+0.01j, 0.9800000000000001-0.01j, 0.99+0.01j, 0.99-0.01j], dtype=complex)

v923z:
> Possible bug with np.all
> 
> ```python
> p = np.array([1j,2j,3j,4j], dtype=np.complex)
> q = np.array([1j,2j,3j,4j], dtype=np.complex)
> print(np.all(p==q))
> 
> >>> False
> ```
> 
> Expect True

I think the problem is that `==` is not implemented for `complex`. I could add that. That is not so expensive, given that the operator is commutative.

I should probably also raise an exception for `<=`, `<`, and similar binary operators, if one of the operands is a complex array.

v923z:
> Possible bug with sorting and conjugate
> 
> ```python
> roots = np.array([0.98-0.01j,0.99+0.01j,0.98+0.01j,0.99-0.01j], dtype=np.complex)
> p = np.sort_complex(roots)
> q = np.sort_complex(np.conjugate(roots))
> print(p)
> print(q)
> 
> >>> array([0.9800000000000001-0.01j, 0.9800000000000001+0.01j, 0.99-0.01j, 0.99+0.01j], dtype=complex)
> >>> array([0.9800000000000001+0.01j, 0.9800000000000001-0.01j, 0.99+0.01j, 0.99-0.01j], dtype=complex)
> ```
> 
> Expect:
> 
> [0.98-0.01j 0.98+0.01j 0.99-0.01j 0.99+0.01j] [0.98-0.01j 0.98+0.01j 0.99-0.01j 0.99+0.01j]

Thanks for reporting the issue! I'll sort it out. (Pun intended.)

wired8:
Hey @v923z your fixes work and I now get the same expected output from a butter filter in ulab as I do in numpy/scipy.  Thank you so much for all you hard work.   I want to contribute my partial implementation of these libraries back into ulab, I'm guessing I add them as snippets?

I was thinking of adding a few libraries under snippets/scipy, snippets/numpy and some tests to boot.  Please advise on how to proceed.

v923z:
> Hey @v923z your fixes work and I now get the same expected output from a butter filter in ulab as I do in numpy/scipy. Thank you so much for all you hard work. 

Not at all. I am glad that it works now. There were a couple of smaller glitches that I discovered after you had reported the problem with `sort_complex`, so your tests were doubly useful. Thanks for sticking with it till the very end!

> I want to contribute my partial implementation of these libraries back into ulab, I'm guessing I add them as snippets?

If the implementation is in `python`, then I would suggest the snippets. If it is in C, then it should be part of [code](https://github.com/v923z/micropython-ulab/tree/master/code). I don't know, whether we can gain anything in speed or RAM, if we implement at least chunks in C. As far as I understand `numpy`, it has helper functions written in C that you don't usually call from `python` directly, only via the `python` implementation of a particular function. We could call them low-level functions that don't have a pythonic interface. I would also be OK with that. If you can point me to the relevant pieces of the `scipy` code, we could discuss what is most sensible path. 

One argument against the C implementation (beyond the obvious) is that it requires flash space, even if you don't use the function. So, snippets are somewhat more flexible.

> I was thinking of adding a few libraries under snippets/scipy, snippets/numpy and some tests to boot. Please advise on how to proceed.

You have just described what to do. I would like to merge this branch, and release 4.0 very soon, but that doesn't really interfere with your stuff. You can PR against this, or the master branch, whenever you are ready. I am quite certain that contributions of this kind would be very useful for the wider community. 

Actually, if you have any comments on the [release draft](https://github.com/v923z/micropython-ulab/releases/tag/untagged-7dcd2e2c0ee810de34c9), please, let me know! 

Also, I think https://github.com/v923z/micropython-ulab/pull/465 is obsolete now. Can we close that?

PR #461, opened 2021-12-17
Change how we alias the float and bool types for mypy

jepler:
mypy 0.920 started giving an error stating
```
ulab/numpy/__init__.pyi:51: error: Cannot assign multiple types to name "_float" without an explicit "Type[...]" annotation
ulab/numpy/__init__.pyi:54: error: Cannot assign multiple types to name "_bool" without an explicit "Type[...]" annotation
```

this quiets the error, and is also compatible with the previous version
(0.910).


v923z:
@jepler Many thanks! For the long run, do you want to make documentation checks part of the CI?

jepler:
It already is.  However, the change which caused the failure was not in ulab or circuitpython, it was in a version of a dependency. So, it only showed up after the new dependency was released AND a build was performed.  The circuitpython philosophy seems to be to pin as little as possible, so this sort of thing just happens from time to time.

v923z:
OK, thanks for the explanation!

PR #471, opened 2022-01-08
implement numpy.save

v923z:
Implements `save` with the 1.0 file format.


v923z:
@labodezao could you please, check, whether this does what you expect?

v923z:
@jepler Jeff, is this the proper (platform-independent) way of writing data to a file? I think the relevant chunks of the code are in https://github.com/v923z/micropython-ulab/blob/68f809801682189f217600b07111d4a0363732fd/code/numpy/io/io.c#L35-L41, and https://github.com/v923z/micropython-ulab/blob/68f809801682189f217600b07111d4a0363732fd/code/numpy/io/io.c#L104

code/numpy/io/io.c
```diff
@@ -0,0 +1,372 @@
...
+            #endif
+            break;
+        #if ULAB_SUPPORTS_COMPLEX
+        case NDARRAY_COMPLEX:
+            #if MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_FLOAT
+            memcpy(buffer+offset, "c8", 2);
+            #else
+            memcpy(buffer+offset, "c16", 2);
+            #endif
+            offset++;
```
jepler, line 292: is this extra increment of offset something that should _only_ happen for c16, not for both c8/c16?

code/numpy/io/io.c
```diff
@@ -0,0 +1,372 @@
...
+    if((bytes_to_read + 51) < header_length) {
+        io_read_(npy, fin, buffer, NULL, header_length - (bytes_to_read + 51), &error);
+    }
+
+    ndarray_obj_t *ndarray = ndarray_new_dense_ndarray(ndim, shape, dtype);
+    char *array = (char *)ndarray->array;
+
+    size_t read = fin->read(npy, array, ndarray->len * ndarray->itemsize, &error);
+    if(read != ndarray->len * ndarray->itemsize) {
+        mp_raise_ValueError(translate("corrupted file"));
```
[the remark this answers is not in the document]
v923z: Good point, thanks for bringing this up!

code/numpy/io/io.c
```diff
@@ -0,0 +1,372 @@
...
+
+#define ULAB_IO_BUFFER_SIZE         128
+
+#define ULAB_IO_NULL_ENDIAN         0
+#define ULAB_IO_LITTLE_ENDIAN       1
+#define ULAB_IO_BIG_ENDIAN          2
+
+#if ULAB_NUMPY_HAS_LOAD
+static void io_read_(mp_obj_t npy, const mp_stream_p_t *fin, char *buffer, char *string, uint16_t len, int *error) {
+    size_t read = fin->read(npy, buffer, len, error);
```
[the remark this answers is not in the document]
v923z: I can make it a bit more verbose, or rename them as `stream` etc.

code/numpy/io/io.c
v923z: Yes, because `offset` is moved forward by two positions in https://github.com/v923z/micropython-ulab/blob/b5d94e8fdcaf31ed60b81ad7303af621d014f3a8/code/numpy/io/io.c#L297, and `c16` is the only `dtype` signature that uses 3 characters. But I have just noticed that https://github.com/v923z/micropython-ulab/blob/b5d94e8fdcaf31ed60b81ad7303af621d014f3a8/code/numpy/io/io.c#L290 is wrong, because we have to compare three characters.

jepler: But it happens for both c8 and c16 as written doesn't it?

v923z: That's right, but I have since corrected the error. I believe 
https://github.com/v923z/micropython-ulab/blob/7672ab9d44ba80168748d777ce83fd65af78cdec/code/numpy/io/io.c#L281-L300 is OK.

docs/manual/source/numpy-functions.rst
```diff
@@ -24,24 +24,26 @@ the firmware was compiled with complex support.
 16. `numpy.interp <#interp>`__
 17. `numpy.isfinite <#isfinite>`__
 18. `numpy.isinf <#isinf>`__
-19. `numpy.max <#max>`__
```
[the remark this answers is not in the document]
v923z: I thought, I have replied to this. In any case, the reason is that the rst file is generated from a jupyter notebook, https://github.com/v923z/micropython-ulab/blob/master/docs/numpy-functions.ipynb. I can run code interactively there, and the output is automatically inserted into the documentation.  But in the notebook, the markdown is exactly as you describe it here, all items in the list are simply prepended with 1., and are re-numerated automatically during the conversion. 

I don't insist on using the notebook, and I would be OK with working with the rst files, if there is a clean way of running `micropython`, or `circuitpython` code there. I believe, those methods are called transformers in sphynx. If you can point me to a half-decent tutorial on how to do this properly, I would appreciate that.

PR #475, opened 2022-01-14
Filter snippets

wired8:
Micropython implementation of Butterworth digital and analog filter with supporting functions.

Code ported from both Scipy and Numpy to run on Micropython.

To run tests:

```bash
bash test-snippets.sh "micropython/ports/unix/micropython-2"
```


snippets/numpy/core/fromnumeric.py
```diff
@@ -0,0 +1,78 @@
...
+from ulab import numpy as np
+from ... import numpy
+
+def prod(arr):
+    result = 1
+    for x in arr:
+        result = result * x
+    return result
+
+def size(a, axis=None):
```
v923z, line 13: I have meant to add this function for a long time, but somehow it slipped off my radar. Do you think it would be a good time now?

snippets/numpy/core/numeric.py
```diff
@@ -0,0 +1,59 @@
+from ulab import numpy as np
+from .multiarray import (asarray)
+
+def zeros_like(a, dtype=None, order='K', subok=True, shape=None):
```
v923z, line 4: We have `zeros`. Is this simply an alias, so that you can deal with the keyword arguments?

snippets/numpy/lib/function_base.py
```diff
@@ -0,0 +1,20 @@
...
+    values = asarray(values)
+    if axis is None:
+        if len(arr.shape) != 1:
+            arr = arr.flatten()
+        values = values.flatten()
+        axis = len(arr.shape)-1
+    return np.concatenate((arr, values), axis=axis)
+
+
+def delete(arr, obj, axis=None):
```
v923z, line 17: `delete` is included with the latest commit.

snippets/numpy/lib/type_check.py
```diff
@@ -0,0 +1,64 @@
+
+from ulab import numpy as np
+from ..core.multiarray import (asarray)
+from ..core.overrides import set_module
+
+@set_module('numpy')
+
+def _isreal(a):
```
v923z, line 8: We can definitely add this function in C. https://numpy.org/doc/stable/reference/generated/numpy.isreal.html

snippets/numpy/core/multiarray.py
```diff
@@ -0,0 +1,21 @@
+from ulab import numpy as np
+
+def asarray(a, dtype=None):
```
v923z, line 3: What is the difference between `asarray`, and `array`? Only that `asarray` returns the array, if an `ndarray` is passed, while `array` returns a new copy? We could definitely add this function in C, if you think it is useful.

v923z:
@wired8 Thanks, this is really fantastic. I would have only one general comment. You have a lot of helper functions, e.g., `asarray`, `isreal` etc., and it would be great to have a top-level module or something like that, where one can easily include these functions without having to go through the complete `snippets` directory. Do you think you could add that?

Later, some of these functions could be implemented in C, but that is for the longer term.

v923z:
@wired8 I have linked 4 issues here that would be closed by merging this. Are those correct?

snippets/numpy/__init__.py
```diff
@@ -0,0 +1,5 @@
+
+from . import core
+from .core import *
+from . import lib
+from .lib import *
```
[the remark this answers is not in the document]
wired8: > @wired8 Thanks, this is really fantastic. I would have only one general comment. You have a lot of helper functions, e.g., `asarray`, `isreal` etc., and it would be great to have a top-level module or something like that, where one can easily include these functions without having to go through the complete `snippets` directory. Do you think you could add that?
> 
> Later, some of these functions could be implemented in C, but that is for the longer term.

Hi @v923z thank you.  When I originally wrote the helper functions the thought was to keep them in 1 file (module), but as they are essentially implementations of Numpy functions I decided it would be better to keep them in the same files as their equivalent Numpy source for easy reference.   I don't believe this is a major problem as the set_module decorator function brings them all under the Numpy namespace.

snippets/numpy/core/fromnumeric.py
wired8: Absolutely.

snippets/numpy/lib/function_base.py
wired8: Will rebase and test with ulab delete.

snippets/numpy/lib/type_check.py
wired8: > @wired8 I have linked 4 issues here that would be closed by merging this. Are those correct?



> @wired8 I have linked 4 issues here that would be closed by merging this. Are those correct?

Yes, this closes all.

snippets/numpy/core/multiarray.py
wired8: My rough implementation of https://numpy.org/doc/stable/reference/generated/numpy.asarray.html

Changing this to C would be much prefered.

snippets/numpy/__init__.py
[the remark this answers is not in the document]
v923z: You are right. Let's leave it at that!

snippets/numpy/core/numeric.py
wired8: The difference is the shape.  https://numpy.org/doc/stable/reference/generated/numpy.zeros_like.html

v923z:
@wired8 Phil, if you want to add a copyright notice to your files before merging, here it is:

```bash
# This file is part of the micropython-ulab project, https://github.com/v923z/micropython-ulab
#
# The MIT License (MIT)
#
# Copyright (c) 2022 Phil Jepsen
```

wired8:
> @wired8 Phil, if you want to add a copyright notice to your files before merging, here it is:
> 
> ```shell
> # This file is part of the micropython-ulab project, https://github.com/v923z/micropython-ulab
> #
> # The MIT License (MIT)
> #
> # Copyright (c) 2022 Phil Jepsen
> ```

Thanks @v923z, added!

snippets/numpy/core/numeric.py
v923z: We could implement it in C, but it wouldn't actually result in speed gains, so I would leave this here.

snippets/numpy/core/multiarray.py
v923z: https://github.com/v923z/micropython-ulab/issues/480

v923z:
@wired8 I am merging this, I have opened two issues https://github.com/v923z/micropython-ulab/issues/479, https://github.com/v923z/micropython-ulab/issues/480 We can update the snippets, once the implementations are in.

v923z:
@wired8 Many thanks for contributing! This is definitely a useful set of functions.

snippets/numpy/core/multiarray.py
v923z: OK, we could add that.

Issue #476, opened 2022-01-14
[FEATURE REQUEST] numpy.lib.format.dtype_to_descr / numpy.lib.format.descr_to_dtype


water5:
`numpy.lib.format.dtype_to_descr`
[https://numpy.org/doc/stable/reference/generated/numpy.lib.format.dtype_to_descr.html](https://numpy.org/doc/stable/reference/generated/numpy.lib.format.dtype_to_descr.html)

`numpy.lib.format.descr_to_dtype`
[https://numpy.org/doc/stable/reference/generated/numpy.lib.format.descr_to_dtype.html](https://numpy.org/doc/stable/reference/generated/numpy.lib.format.descr_to_dtype.html)

v923z:
@water5 I think I would handle this in `snippets`. Could you, please, check out https://github.com/v923z/micropython-ulab/tree/json, and try the script in https://github.com/v923z/micropython-ulab/tree/json/snippets. 

This works with `hexlify` at the moment. If you are OK with this, I can add the reverse function with `unhexlify`. Unfortunately, `a2b_base64` does not work as expected. I have raised an issue: https://github.com/micropython/micropython/issues/8175

water5:
> This works with hexlify at the moment. If you are OK with this, I can add the reverse function with unhexlify. Unfortunately, a2b_base64 does not work as expected. I have raised an issue: micropython/micropython#8175

You should use `ubinascii.b2a_base64`?
```python
from ubinascii import b2a_base64, a2b_base64
b2a_base64(bytearray([1, 2, 3]))
```
> b'AQID\n'

And `CPython` side, between `base64.b64encode` and `binascii.a2b_base64` have little difference.
`CPython` and `MicroPython` two side should use same library, but `MicroPython` only implement `ubinascii`.

-----------

Notice:
On `MicroPython`, `ubinascii.b2a_base64` including a newline char at the end by default, and can't choose by parameter.
[https://docs.micropython.org/en/latest/library/binascii.html#binascii.b2a_base64](https://docs.micropython.org/en/latest/library/binascii.html#binascii.b2a_base64)

On `CPython`, have `newline` parameter, `newline=True` by default.
[https://docs.python.org/3/library/binascii.html#binascii.b2a_base64](https://docs.python.org/3/library/binascii.html#binascii.b2a_base64)

v923z:
[...]

PR #477, opened 2022-01-14
json snippets

v923z:
This PR adds two snippets, with which one can convert ndarrays to serialised json string, and vice versa. This should address the issue mentioned in https://github.com/v923z/micropython-ulab/issues/468#issuecomment-1012372905 and thereafter, and should also close https://github.com/v923z/micropython-ulab/issues/476.


v923z:
@jepler We don't need a review here, since this doesn't concern the core, everything is in python, but it might be interesting for the `circuitpython` community. The same applies to https://github.com/v923z/micropython-ulab/pull/475.

PR #481, opened 2022-01-14
add implementation of size and asarray

v923z:


v923z:
@wired8 Phil, here is the implementation of `asarray`, and `size`. Could you,please, check it out? My tests have passed, but that might not mean too much. If these are OK, then I would merge it as soon as possible, so that you can update your snippets. Thanks!

wired8:
When calling asarray in micropython:
```python
>>> np.asarray(5)
TypeError: wrong input type
>>> np.asarray([5])
ValueError: bad typecode
```

Regular python:
```python
print(np.asarray(5))
5
print(np.asarray([5]))
[5]
```

Note:  asarray will always return a type of ndarray.

https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects

v923z:
@wired8 Thanks for catching this. I have updated the code. If you think that the test script have to be extended, so be it.

wired8:
Also np.size:

Micropython:
```python
>>> np.size(5)
TypeError: first argument must be an ndarray
>>> np.size([5])
TypeError: first argument must be an ndarray
```

Regular python:
```python
print(np.size(5))
1
print(np.size([5]))
1
```

Regular python accepts an array_like.

v923z:
@wired8 I believe it should be OK now. Let me know if there are still glitches.

wired8:
Breaks numpy.any

Actual:
```python
Wn = 5
print(np.any(Wn <= 0))
TypeError: 'bool' object isn't iterable
```


Expected:

```python
Wn = 5
print(np.any(Wn <= 0))
False
```

v923z:
> Breaks numpy.any
> 
> Actual:
> 
> ```python
> Wn = 5
> print(np.any(Wn <= 0))
> TypeError: 'bool' object isn't iterable
> ```

That is actually a totally separate issue. And I am not sure, whether it is worth the overhead. Virtually all `numpy` functions should be able to handle scalars, and the only way of handling this is adding an `if` clause at the beginning of each function, and if the input is a scalar, then return there, otherwise, move to the array handling section. My guess is that this adds something like 50-100 bytes for each function, so overall, we are talking about 5-10 kB extra. `circuitpython` is already very tight on the flash, and I would rather drop extra features and corner cases than complete functions.

There is a reason, why `numpy` occupies 30+ MB of space, while `ulab` only approx. 100 kB. I am not saying that we shouldn't strive for compatibility, but it doesn't come cheap. I am also open to discussion. I am merging this as is, but if you think that this issue is really troublesome, open a separate ticket, and we can pick it up from there. This is definitely a worthwhile discussion, and I am glad you brought this up.

Issue #435, opened 2021-08-25
[BUG] Spectrogram is not like scipy's spectrogram?


jepler:
```python
>>> import ulab
>>> ulab.__version__
'3.3.1-2D'
```

**To Reproduce**
Place the following in a file and run it with micropython & python3:
```python
try:
  from ulab import numpy as np
except ImportError:
  import numpy as np
  
try:
  from ulab.scipy.signal import spectrogram
except ImportError:
  from scipy.signal import spectrogram

data = np.array([1,2,1,4])
print(spectrogram(data))
```

However, the return types of the spectrogram functions are entirely different, and the values don't seem related in an obvious way.
```
$ python -Wignore yon2.py
(array([0.  , 0.25, 0.5 ]), array([2.]), array([[0.33333333],
       [3.33333333],
       [3.        ]]))
$ ./micropython-coverage yon2.py
array([8.0, 2.0, 4.0, 2.0], dtype=float64)
```

**Expected behavior**
I wanted to create a program that finds the spectrum of some signal and is compatible between scipy and ulab, but the interfaces don't seem to be the same.

**Additional context**
Add any other context that might help to locate the root of the problem.

v923z:
So, if I understand the issue properly, the problem is that the meaning of `spectrogram` is different in `scipy`, and `ulab`. I have just searched for that, and it seems to me that `signal.welch` is the closest to what is needed here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.welch.html, or, perhaps, `matplotlib.mlab.psd`: https://www.kite.com/python/docs/matplotlib.mlab.psd

I see four possibilities for a fix; none is ideal:
1. we remove the `spectrogram` function; that re-establishes compatibility, but introduces RAM overhead for an operation that is almost always required. Frankly, I think the power spectrum is probably one of the most-used functions, and I am pretty sure that it is more useful than the FFT itself. 
1. we move the spectrogram function `utils`. That also re-established compatibility, but keeps the function. The downside is that it is not clear what should happen on a PC.
1. we rename/rework the current implementation, and introduce `welch` instead. I think `welch` and the current implementation of  `spectrogram` are identical, if you take a single (non-overlapping) signal segment.
1. we move the function to a new module, `mlab`, and re-brand it as `psd`. This might make sense, if there are other functions of the `mlab` module that would be beneficial. 

[...]

PR #499, opened 2022-01-29
move spectrogram to utils module

v923z:
This PR should resolve the incompatibility issue mentioned in https://github.com/v923z/micropython-ulab/issues/435. Since there is no similar function in either `scipy`, or `numpy`, `spectrogram` is moved to the `numpy`-incompatible `utils` module.


docs/manual/source/conf.py
```diff
@@ -27,7 +27,7 @@
 author = 'Zoltán Vörös'
 
 # The full version, including alpha/beta/rc tags
-release = '4.4.0'
+release = '4.4.1'
```
jepler, line 30: Do you use semantic versioning? If so, a move like this would call for a major version bump.

docs/manual/source/scipy-signal.rst
```diff
@@ -6,7 +6,6 @@ Functions in the ``signal`` module can be called by prepending them by
 ``scipy.signal.``. The module defines the following two functions:
```
jepler, line 6: It now defines a single function, yes?

jepler (changes requested):
A few notes, one thing in the doc should definitely be fixed

docs/manual/source/scipy-signal.rst
v923z: That's right. I have updated the docs.

docs/manual/source/conf.py
v923z: Definitely true. Thanks for catching this!

jepler (approved):
Thanks for addressing my suggestions.  No other testing performed.

PR #505, opened 2022-02-10
fix complex slicing

v923z:


wired8:
Looks good :+1:, thanks @v923z

PR #508, opened 2022-02-19
Properly register submodules of ulab

jepler:
This is related to
 * https://github.com/adafruit/circuitpython/issues/6066

in which, after the merge of 1.18 into CircuitPython, we lost the ability to import submodules of built-in modules.

While reconstructing the changes we had made locally to enable this, I discovered that there was an easier way: simply register the dotted module names via MP_REGISTER_MODULE.

In CircuitPython, we'd like this as a hotfix for ulab 4.0.0, so instead of just merging this to master we may need to create a branch for it (or re-use "legacy", which was previously at 1.7.x I think also for CircuitPython)

This is not absoutely urgent, it does not affect the upcoming 7.2.0 release of CircuitPython, just our 'main' development branch where we merged MicroPython 1.18.

### DIFF
--- a/build-cp.sh
+++ b/build-cp.sh
@@ -35,7 +35,7 @@ readlinkf_posix() {
   done
   return 1
 }
-NPROC=$(python -c 'import multiprocessing; print(multiprocessing.cpu_count())')
+NPROC=$(python3 -c 'import multiprocessing; print(multiprocessing.cpu_count())')
 HERE="$(dirname -- "$(readlinkf_posix -- "${0}")" )"
 [ -e circuitpython/py/py.mk ] || (git clone --no-recurse-submodules --depth 100 --branch main https://github.com/adafruit/circuitpython && cd circuitpython && git submodule update --init lib/uzlib tools)
 rm -rf circuitpython/extmod/ulab; ln -s "$HERE" circuitpython/extmod/ulab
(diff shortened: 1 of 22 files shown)


jepler:
weeeeeird, on circuitpython this gives module names like `ulab.numpy` (can be seen in `help("modules")` but in micropython it gives mdule names like `ulab_dot_numpy`.  Turns out, this ability to have 'escape-like' characters interpreted in qstrs is a circuitpython superpower.

v923z:
> While reconstructing the changes we had made locally to enable this, I discovered that there was an easier way: simply register the dotted module names via MP_REGISTER_MODULE.
> 
> In CircuitPython, we'd like this as a hotfix for ulab 4.0.0, so instead of just merging this to master we may need to create a branch for it (or re-use "legacy", which was previously at 1.7.x I think also for CircuitPython)

I think, a meaningful point of branching off would be 88231bc3a1e5341dc8e1f2d2ecbda50be57cae67 That contains a couple of new functions on top of 4.0.0, and a number of fixes for issues that popped up on the way. The next commit is 5.0.0, and I think you don't want that at the moment. That contains the breaking change, which moves the `spectrogram` function to `utils`.

v923z:
> weeeeeird, on circuitpython this gives module names like `ulab.numpy` (can be seen in `help("modules")` but in micropython it gives mdule names like `ulab_dot_numpy`. Turns out, this ability to have 'escape-like' characters interpreted in qstrs is a circuitpython superpower.

This is actually an interesting point, and also came up in context of the `loadtxt` function. One of the keyword arguments should be ';', I believe, and I couldn't implement that. 

Is it correct that MP_QSTRs are converted by some `python` script, before the pre-processor has a change of inspecting the code? In other words, one could add this functionality without having to tamper with the core, and adding a single byte to the firmware, couldn't one?

PR #522, opened 2022-04-20
Update reference to spectrum

tekktrik:


v923z:
Thanks for the PR! We removed `spectrum`, because there were no similar functions in `numpy`, or `scipy`, and moved the functionality to `utils.spectrogram`: https://github.com/v923z/micropython-ulab/blob/10b79f2041065348b600cf9b83f177f204b32f9a/code/utils/utils.c#L206. Since `circuitpython` wants to pick up `master` with release 8.0, I would suggest that the documentation point to that location.

tekktrik:
Will do, thanks!

v923z:
@tekktrik Alec, the function is bound to `spectrogram` in the `python` name space. I am really not against changing this to `spectrum`, if that is what you want, however, if we do so, I would like the function in C to be called `spectrum`, so that the code is aligned with the documentation.

tekktrik:
Oh wow, no I just didn't realize it was a typo haha.  Whoopsie, I'll fix that

v923z:
@tekktrik https://github.com/v923z/micropython-ulab/runs/6131134662?check_suite_focus=true fails on the `circuitpython` documentation. You can actually debug this locally, if you run the https://github.com/v923z/micropython-ulab/blob/master/build-cp.sh script from the root of the repository.

tekktrik:
Will do, thanks for all the help!

v923z:
@tekktrik Again, many thanks for contributing!

Issue #528, opened 2022-06-10
Unable to install micropython for rp2, cmake issues


MicahHaack:
I have been trying to follow the instructions for installing micropython for the rp2 (using ubuntu on windows subsystem), using the provided bash file found here: (https://github.com/v923z/micropython-ulab/blob/master/build/rp2.sh)

It works well until the last couple commands for cmake, where I get the following errors:

![image](https://user-images.githubusercontent.com/44842363/173094668-842b4598-c017-4c4d-922c-b08c7af48d51.png)

I have installed all build essentials, cmake, make, gcc, and g++, and all are up to date. I'm not sure how to fix this.

v923z:
I noticed that the latest version of `micropython` throws the compile process off. As a temporary fix, you can roll back your local `micropython` version by two, or three days. I'll try to find out, what causes the issue.

v923z:
@Gadgetoid Philip, do you happen to know what is going on here?

Gadgetoid:
This just isn't finding "make" or GCC- I'll try and run a build with the latest version and see what happens.

v923z:
https://github.com/v923z/micropython-builder didn't produce artifact in the latest run, and that's why I think that something is amiss. I just haven't had time to find out the details.

Gadgetoid:
Completely *different* error. Good fun:

```
<command-line>: error: expected identifier or '(' before numeric constant
/home/phil/Development/pico/micropython-latest/ports/rp2/build-PICO/genhdr/moduledefs.h:87:57: note: in expansion of macro 'MODULE_ULAB_ENABLED'
   87 | extern const struct _mp_obj_module_t ulab_user_cmodule, MODULE_ULAB_ENABLED;
      |                                                         ^~~~~~~~~~~~~~~~~~~
/home/phil/Development/pico/micropython-latest/py/objmodule.c:169:5: error: macro "MP_ROM_PTR" passed 2 arguments, but takes just 1
  169 |     MICROPY_REGISTERED_MODULES
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/phil/Development/pico/micropython-latest/py/mpstate.h:35,
                 from /home/phil/Development/pico/micropython-latest/py/runtime.h:29,
                 from /home/phil/Development/pico/micropython-latest/py/bc.h:30,
                 from /home/phil/Development/pico/micropython-latest/py/objmodule.c:32:
[...]

PR #529, opened 2022-06-10
Drop third argument from MP_REGISTER_MODULE.

Gadgetoid:
As noticed in #528 a recent change to MicroPython upstream has broken module builds.

This is because the third argument of `MP_REGISTER_MODULE` has been dropped.

This fix should *maybe* be merged to coincide with the next release of MicroPython which will no doubt ratify and document this change, but  I'll leave this up to you.

I've adopted the pattern from https://github.com/micropython/micropython/blob/efe23aca7154d21a9e8f283313c7a1ac29e05d86/docs/develop/cmodules.rst

The breaking commit to MicroPython is: https://github.com/micropython/micropython/commit/efe23aca7154d21a9e8f283313c7a1ac29e05d86


v923z:
@Gadgetoid Thanks for the PR! This fails with `circuitpython`, but I can fix that later today.

v923z (commented):
Could you, please, set 

#define ULAB_VERSION 5.0.8

Gadgetoid:
A little more digging and it looks like we could check `MICROPY_VERSION` to pick the correct `MP_REGISTER_MODULE` format depending on which version of MicroPython it's being built against.

tannewt:
> @tannewt @dhalbert when do you plan to pick up [micropython/micropython@efe23ac](https://github.com/micropython/micropython/commit/efe23aca7154d21a9e8f283313c7a1ac29e05d86)?

We follow MP releases. So, after 1.19 is released. How soon after will depend on what we're up to. (Probably after we release 8 since 1.19 will require us to release 9.x.)

Gadgetoid:
It wont help with your CI but I've had a fix merged into MicroPython that corrects this loud explosion to a specific error: https://github.com/micropython/micropython/pull/8759

I think running CI against the bleeding edge of MicroPython may not be considered best practice, but it's actually been enormously useful in this instance for notifying me what's going to explode!

v923z:
> I think running CI against the bleeding edge of MicroPython may not be considered best practice, but it's actually been enormously useful in this instance for notifying me what's going to explode!

I definitely see your point, and I think there is no ideal solution to this problem. However, aligning with the latest gives people the opportunity to test the bleeding edge. I would also assume that `micropython` improves with each commit, so the latest is going to be better than the last release.

v923z:
@Gadgetoid Philip, many thanks for your efforts! I think this is a satisfactory solution, and I would just leave it at that. Other parties that depend on `micropython` and `ulab`, e.g. `circuitpython`, `openmv` and the like, usually pin the commit pointer, so this should be OK.

PR #532, opened 2022-07-06
Fix error message for maximum number of dimensions

dpgeorge:
Prior to this change, if you built ulab with the default 2 dimensions and ran code that needed more than 2 (eg `tests/4d/complex/imag_real.py`) it would error out with the confusing message "ValueError: maximum number of dimensions is 4".


v923z:
> Prior to this change, if you built ulab with the default 2 dimensions and ran code that needed more than 2 (eg `tests/4d/complex/imag_real.py`) it would error out with the confusing message "ValueError: maximum number of dimensions is 4".

Indeed. Many thanks!

v923z:
Again, thanks for your contribution, comments, and insights!

dpgeorge:
Thanks for merging!

PR #533, opened 2022-07-07
Fix several build errors in CircuitPython

jepler:
These errors actually occurred when building the raspberry pi pico port; they weren't deteted when building the unix port during ulab ci, as this does not enable the diagnostics -Werror=undefined or -Werror=discarded-qualifiers.


v923z:
Thanks, Jeff!

PR #535, opened 2022-07-11
fix use of MICROPY_FLOAT_CONST, and build single-precision as part of CI

dpgeorge:
Follow on from #534.


dpgeorge:
This removes the `+0.` added in #534 and replaces it with a more long-term solution.

Also now builds a single precision port as part of CI.

v923z:
@dpgeorge Thanks for cleaning this up!
@jepler You beat me to merging this, thanks a lot!

PR #536, opened 2022-07-11
Make optimize's bisect,fmin,newton use C stack instead of heap

dpgeorge:
This is a simple change to reduce the use of heap allocation in `optimize.bisect`, `optimize.fmin` and `optimize.newton`.

### DIFF
--- a/code/scipy/optimize/optimize.c
+++ b/code/scipy/optimize/optimize.c
@@ -74,7 +74,7 @@ STATIC mp_obj_t optimize_bisect(size_t n_args, const mp_obj_t *pos_args, mp_map_
         mp_raise_TypeError(translate("first argument must be a function"));
     }
     mp_float_t xtol = mp_obj_get_float(args[3].u_obj);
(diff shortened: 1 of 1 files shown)


v923z:
@dpgeorge Thanks! I see that my use of the heap is a bit sloppy and ad-hoc. I'll try to watch out for this in the future.

dpgeorge:
I did review all the memory allocations in ulab and these were the ones that could be eliminated.

v923z:
Actually, now that you mention it, things like https://github.com/v923z/micropython-ulab/blob/2b4292abcc4f22acf242b2b4656a65cd124f1a5e/code/ndarray.c#L910 could be taken out, because `ULAB_MAX_DIMS` is known at compile time. I don't see any reason for dynamically allocating the memory. N.B., my comment is not a hint at anything.

dpgeorge:
Yes, that does look like a temporary `shape` array that can be allocated on the C stack.

v923z:
OK, thanks for the confirmation! I'll try to weed these instances out.

Issue #541, opened 2022-07-26
circuitpython and micropython tests deliver differing results


v923z:
https://github.com/micropython/micropython/commit/f9cbe6bc47dd4f5b8e85178caecd6f0de22b4c34 modifies the formatting of certain floats, while `circuitpython` keeps the old implementation. This difference leads to differing outputs on the test scripts.

@jepler @tannewt @dhalbert What would be the preferable fix for this till `circuitpython` pulls in the new implementation? I would not like to have two tiers of tests, but I also undrestand that `ulab` should be tested against `circuitpython`, too.

@jepler By assigning the issue to you I don't want to imply that you have to fix it, but I wanted to have someone from that side.

dhalbert:
I think we could cherry-pick the formatting fix for 8.0.0, and ignore it for 7.3.x. It will not be an issue until we try to update ulab.

v923z:
@dhalbert Thanks for the feedback! I could take it out of the CI for the time being. We just have to make sure that we put it back, once you merged the float formatting. Would it be possible to add this to the 8.0.0 milestone, so that we don't forget about it?

dhalbert:
Oh, I didn't realize it was being tested in your CI. I wonder if you can make the tests be rather more forgiving (limiting significant digits), or are there formatting changes all over the place?

v923z:
This issue pops up, whenever the result of whatever operation is an integer float. There are several test scripts that produce that. Here is an example: https://github.com/v923z/micropython-ulab/runs/7528072036?check_suite_focus=true

dhalbert:
In that example, is `988.0000000000001` the value that CircuitPython currently produces or that the newly updated MicroPython produces?

v923z:
This is the value produced by `circuitpython`. `micropython` now prints just `998.0`.

v923z:
I haven't looked at the implementation in detail, but the commit message is 
```
Formerly, py/formatfloat would print whole numbers inaccurately with
nonzero digits beyond the decimal place.  This resulted from its strategy
of successive scaling of the argument by 0.1 which cannot be exactly
represented in floating point. ```

dhalbert:
The new fix is a good one, I just wanted to double-check.

dpgeorge:
In evaluating that improvement to formatfloat, I did run it on the ulab test suite.  IMO the output of ulab tests is much better with this formatfloat fix.

v923z:
[...]

PR #542, opened 2022-07-26
fix: remove make axtls, added make submodules

tejalbarnwal:
There has been an update with the new build instructions for micropython as found [here](https://github.com/micropython/micropython#the-unix-version)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -45,7 +45,7 @@ HERE="$(dirname -- "$(readlinkf_posix -- "${0}")" )"
 [ -e micropython/lib/axtls/README ] || (cd micropython && git submodule update --init lib/axtls )
 dims=${1-2}
 make -C micropython/mpy-cross -j${NPROC}
-make -C micropython/ports/unix -j${NPROC} axtls
+make -C micropython/ports/unix submodules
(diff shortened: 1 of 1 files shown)


v923z:
Thanks for the comments! I remove `axtls` from the build script. https://github.com/v923z/micropython-ulab/pull/540 passes with `micropython`. We are working on the `circuitpython` CI, but that is a different issue: https://github.com/v923z/micropython-ulab/issues/541

v923z:
What I wanted to say is that the other PR should actually contain the fix for this.

tejalbarnwal:
Just to confirm, I ran `build.sh` while commenting the line to run the test scripts. I am referring to the below code line
```bash
bash test-common.sh "${dims}" "micropython/ports/unix/micropython-$dims"
```
The above line interrupts the build process because `gc.py` fails the test case.
image

The `gc.py.exp` is as follows
```
988.0000000000001
array([1.0, 1.0, 1.0, ..., 1.0, 1.0, 1.0], dtype=float64)
988.0000000000001
array([1.0, 1.0, 1.0, ..., 1.0, 1.0, 1.0], dtype=float64)
```
`gc.py.out` gives the following:
```
988.0
array([1.0, 1.0, 1.0, ..., 1.0, 1.0, 1.0], dtype=float64)
988.0
array([1.0, 1.0, 1.0, ..., 1.0, 1.0, 1.0], dtype=float64)
```
Am not sure why does it fails the test

v923z:
Again, the CI produces no errors on `micropython`: https://github.com/v923z/micropython-ulab/runs/7528073593?check_suite_focus=true

tejalbarnwal:
Right! Thank you

v923z:
I think this PR is no longer needed.

PR #544, opened 2022-07-29
Added test script for np.nonzero()

tejalbarnwal:
added following script at `ulab/tests/2d/numpy/nonzero.py`
```python
try:
   from ulab import numpy as np
except:
   import numpy as np

array = np.array(range(16)).reshape((4,4))
boolean_array = array < 3

# np.nonzero() returns a tuple of corresponding row and column 
# indices array for each element being true in boolean array
row_indices, column_indices = np.nonzero(boolean_array)

print("Given Array:\n", array)
print("\nBoolean Array:\n", boolean_array)
print("\nFollowing indices have non-zero values:")
for i in range(len(row_indices)):
	print("row", row_indices[i], "and column", column_indices[i])
```
The expected output file is located at `ulab/tests/2d/numpy/nonzero.py.exp`
Expected output:
```
Given Array:
 array([[0.0, 1.0, 2.0, 3.0],
       [4.0, 5.0, 6.0, 7.0],
       [8.0, 9.0, 10.0, 11.0],
       [12.0, 13.0, 14.0, 15.0]], dtype=float64)

Boolean Array:
 array([[True, True, True, False],
       [False, False, False, False],
       [False, False, False, False],
       [False, False, False, False]], dtype=bool)

Following indices have non-zero values:
row 0 and column 0
row 0 and column 1
row 0 and column 2
```

### DIFF
--- a/tests/2d/numpy/nonzero.py
+++ b/tests/2d/numpy/nonzero.py
@@ -1,0 +1,17 @@
+try:
+   from ulab import numpy as np
+except:
+   import numpy as np
+
+array = np.array(range(16)).reshape((4,4))
+boolean_array = array < 3
+
+# np.nonzero() returns a tuple of corresponding row and column 
+# indices array for each element being true in boolean array
+row_indices, column_indices = np.nonzero(boolean_array)
+
+print("Given Array:\n", array)
+print("\nBoolean Array:\n", boolean_array)
+print("\nFollowing indices have non-zero values:")
+for i in range(len(row_indices)):
+	print("row", row_indices[i], "and column", column_indices[i])
--- a/tests/2d/numpy/nonzero.py
+++ b/tests/2d/numpy/nonzero.py
@@ -4,14 +4,13 @@ except:
    import numpy as np
 
 array = np.array(range(16)).reshape((4,4))
-boolean_array = array < 3
+print(array)
+print(array < 5)
+print(np.nonzero(array < 5))
 
-# np.nonzero() returns a tuple of corresponding row and column 
-# indices array for each element being true in boolean array
-row_indices, column_indices = np.nonzero(boolean_array)
+dtypes = (np.uint8, np.int8, np.uint16, np.int16, np.float)
 
-print("Given Array:\n", array)
-print("\nBoolean Array:\n", boolean_array)
-print("\nFollowing indices have non-zero values:")
-for i in range(len(row_indices)):
-	print("row", row_indices[i], "and column", column_indices[i])
+for dtype in dtypes:
+    array = (np.arange(2, 12, 3, dtype=dtype)).reshape((2,2)) - 2
+    print(array)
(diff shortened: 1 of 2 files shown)


v923z:
`nonzero` works with arrays of arbitrary `dtype`, so e.g., this is also valid:

```python
>>> a = np.array(range(9), dtype=np.uint8).reshape((3,3)) - 4
>>> print(np.nonzero(a))
```
You can simply loop through the types like in https://github.com/v923z/micropython-ulab/blob/master/tests/2d/numpy/arange.py. Would you mind adding this, so that the test would be complete?

tejalbarnwal:
sure, I will update with this

v923z:
> sure, I will update with this

OK, thanks! Let me know when are ready. The rest of the PR is done, so as soon as you can update your test script, we could merge this into master.

tests/2d/numpy/nonzero.py
```diff
@@ -0,0 +1,17 @@
+try:
+   from ulab import numpy as np
+except:
+   import numpy as np
+
+array = np.array(range(16)).reshape((4,4))
+boolean_array = array < 3
+
+# np.nonzero() returns a tuple of corresponding row and column
```
v923z, line 9: Don't leave trivial comments in the test script! These are not even for humans, and there is the documentation, anyway.

tests/2d/numpy/nonzero.py
```diff
@@ -0,0 +1,17 @@
...
+boolean_array = array < 3
+
+# np.nonzero() returns a tuple of corresponding row and column 
+# indices array for each element being true in boolean array
+row_indices, column_indices = np.nonzero(boolean_array)
+
+print("Given Array:\n", array)
+print("\nBoolean Array:\n", boolean_array)
+print("\nFollowing indices have non-zero values:")
+for i in range(len(row_indices)):
```
v923z, line 16: You really don't need to loop through rows and columns. This is not a demonstration script, so you can just print out the results as in https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html.

tests/2d/numpy/nonzero.py
tejalbarnwal: Hey, I removed all the comments

tests/2d/numpy/nonzero.py
tejalbarnwal: I removed this too

tejalbarnwal:
I have updated the script to loop through different data types

tests/2d/numpy/nonzero.py
v923z: Thanks!

v923z:
Fantastic! Many thanks for sticking with it till the very end! I hope you enjoyed the ride, and I am looking forward to your future contributions.

tejalbarnwal:
Yes, sure

PR #550, opened 2022-09-29
Minor compile fixes: comma and new-line at end of files

dpgeorge:
Two commits here which aim to fix compile errors on older systems:
- Fix a missing comma when using old versions of MicroPython.
- Some very old compilers don't like files without a new-line at the end.


dpgeorge:
With this patch I managed to get ulab building and running the tests on MicroPython v1.9.3 with a gcc v4.2.1.

v923z:
Thanks a lot! I didn't even know there was an issue with the compiler.

However, this seems to fail with the mac. As far as I understand, there is a glitch with the float formatting. If you can fix it, great, if you can't, then I would just remove that particular entry from the test.

dpgeorge:
I fixed the mac test.

There's still a problem with the CircuitPython build, it has a warning with auto-generated markdown.  That looks like it's impossible to fix here.

v923z:
> I fixed the mac test.

Thanks a lot, I really appreciate your efforts!

> There's still a problem with the CircuitPython build, it has a warning with auto-generated markdown. That looks like it's impossible to fix here.

I think I would ignore that for now. I'll ping Jeff.

v923z:
@jepler Jeff, it seems to me that `cricuitpython` modified the CI workflow files, and some of these changes didn't propagate to `ulab`, so the documentation generation fails now. Could you, please, take a look at it?

dpgeorge:
Thanks for merging!

jepler:
Thanks! I believe I've fixed the doc build problem by a change in CircuitPython, part of another PR: https://github.com/adafruit/circuitpython/pull/6960/commits/325cbb156490022b264581acb1fe6f5abaef0fc9

I did a re-run of the failed jobs in the workflow on the tip of master and it passed now: https://github.com/v923z/micropython-ulab/actions/runs/3157441720

Issue #551, opened 2022-10-09
[BUG]:  fft.ifft gives incorrect results with complex support


A622266:
When printing a complex array, an exception is thrown and the program terminates.
The following code runs correctly in CPython.  It also appears to run correctly in ulab on my esp32  if the print(z) statement is omitted.  If print(z) is replaced by print(t), the array is output as expected, and matches the CPython output for the array t.  The output when print(z) is run is included below, showing ulab version and error message.

```python
try:
    import ulab
    from ulab import numpy as np
    from ulab import scipy as spy
    print('version string: ', ulab.__version__)
    print('you are running ulab')

except ImportError:
    import numpy as np
    import scipy as spy
    import matplotlib.pyplot as plt
    print('you are runnning CPython')

def analytic_signal(x):
     N = len(x)
     X = np.fft.fft(x)
     h = np.zeros(N)
     h[0] = 1
     h[1:N//2] = 2*np.ones(N//2-1)
     h[N//2] = 1
     Z = X*h
     z = np.fft.ifft(Z)
     return z

# from analytic_signal import analytic_signal

t = np.linspace(0.0, 0.01, num=2**7)
x = np.sin(2*np.pi*700*t)
z = analytic_signal(x)
print(z)

try:
    plt.subplot(2,1,2)
    plt.plot(t,z.real,'k',label='Real(z[n])')
    plt.plot(t,z.imag,'r',label='Imag(z[n])')
    plt.title('Components of Analytic signal')
    plt.xlabel('n')
    plt.ylabel('z_r[n] and z_i[n]')
    plt.legend()
    plt.show()

except NameError:
    pass
```

Here is the output:
<<START OUTPUT>>
version string:  6.0.0-4D-c
you are running ulab
Guru Meditation Error: Core  0 panic'ed (InstrFetchProhibited). Exception was unhandled.
Core 0 register dump:
PC      : 0xffffffff  PS      : 0x00060e30  A0      : 0x8013e779  A1      : 0x3ffcde80  
A2      : 0x3ffcdeac  A3      : 0x3ffe7590  A4      : 0x00000000  A5      : 0x00000000  
A6      : 0x3f41a138  A7      : 0x3ffcdea0  A8      : 0x80115a70  A9      : 0x3ffcde60  
A10     : 0x3ffcdeac  A11     : 0x3ffe7590  A12     : 0x00000000  A13     : 0x3ffcdee0  
[...]

PR #557, opened 2022-11-05
fix segmentation fault bug in fft.ifft

v923z:
Should fix the bug reported in https://github.com/v923z/micropython-ulab/issues/551


v923z:
@jepler Jeff, do we have to be concerned by https://github.com/v923z/micropython-ulab/actions/runs/3399781965/jobs/5653757709, or can we just go ahead, and merge the fix? I don't quite understand why the CI fails.

jepler:
`functools.cache`is only in python 3.9. can you bump the version used during CI to 3.9 or 3.10? It should fix the problem

PR #563, opened 2022-11-29
Fix sort when dtype is uint16

dpgeorge:
Prior to this fix the code was using the mp_float_t data type for uint16 and producing incorrect sort results.


v923z:
Many thanks! Merged.

dpgeorge:
Thanks for the quick merge!

PR #577, opened 2023-01-15
Allocate 1d results of correct size

### LINKED ISSUE #574: [BUG] Memory corruption with ulab.numpy.dot
**Describe the bug**
```
Adafruit CircuitPython 8.0.0-beta.6 on 2022-12-21; Adafruit ItsyBitsy M4 Express with samd51g19
>>> ulab.__version__
'6.0.2-2D-c'
```
as well as "unix circuitpython" with the same ulab version

**To Reproduce**
```python
import ulab.numpy as np
coeff = np.ones((6,6))
values = np.ones(6) 
# next line prints array([6.0]) in ulab while python3 prints [6. 6. 6. 6. 6. 6.] 
print(np.dot(values, coeff))
# eventually causes crash!
while True: x = np.dot(values, coeff)
```

**Expected behavior**
Would like behavior to match standard Python & not hardfault

**Additional context**
Is an output array of wrong size being allocated?

jepler:
Before, it was erroneously allocated as a 1-element array instead of a 6-element array in the test case.

Closes #574

### DIFF
--- a/code/numpy/transform.c
+++ b/code/numpy/transform.c
@@ -380,7 +380,7 @@ mp_obj_t transform_dot(mp_obj_t _m1, mp_obj_t _m2) {
     if(ndim == 2) { // matrix times matrix -> matrix
         shape = ndarray_shape_vector(0, 0, shape1, shape2);
     } else { // matrix times vector -> vector, vector times vector -> vector (size 1)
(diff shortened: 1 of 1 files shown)


v923z:
Thanks!

Issue #581, opened 2023-01-22
[BUG] sorting of complex arrays causes crash


v923z:
As mentioned in https://github.com/v923z/micropython-ulab/issues/579, `sort_complex` results in a crash.

```python

try:
    from ulab import numpy as np
except:
    import numpy as np

dtypes = (np.uint8, np.int8, np.uint16, np.int16, np.float, np.complex)

for dtype in dtypes:
    print(np.sort_complex(np.array(range(5, 0, 81), dtype=dtype)))

```

PR #583, opened 2023-01-22
fix sorting of empty  arrays in sort_complex

v923z:
Should fix https://github.com/v923z/micropython-ulab/issues/581


jepler (approved):
Looks sensible, thanks

PR #591, opened 2023-03-03
Use enum instead of const for float consts with repr C

dpgeorge:
Older compilers (eg for esp8266) cannot see that "const uint32_t" is a true constant at compile time, so use an assigned enum instead.

Don't do this for object representation D because that requires a 64-bit integer.


v923z:
Many thanks, I really appreciate these contributions, because I would never find such nuances.

dpgeorge:
You're welcome!

PR #602, opened 2023-04-24
Add a ULAB_PRINT_DTYPE constant

HugoNumworks:
I don't know if it is something you would like to add, but I added a small option for the display of "dtype=..." when printing an array.

### DIFF
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -525,6 +525,7 @@ void ndarray_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t ki
         ndarray_print_bracket(print, 0, self->shape[ULAB_MAX_DIMS-4], "]");
         #endif
     }
+    #if ULAB_PRINT_DTYPE
     mp_print_str(print, ", dtype=");
     if(self->boolean) {
         mp_print_str(print, "bool)");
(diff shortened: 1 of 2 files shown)


v923z:
> > there seems to be some discrepancy in `numpy` itself
> 
> I don't use Numpy very often, but I think I have never seen the dtype printed alongside the array, in which case does it happen ?

This doesn't print the `dtype`: 

```python
>>> linspace(0, 10, num=11)
array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])
```

HugoNumworks:
> > > there seems to be some discrepancy in `numpy` itself
> > 
> > 
> > I don't use Numpy very often, but I think I have never seen the dtype printed alongside the array, in which case does it happen ?
> 
> This doesn't print the `dtype`:
> 
> ```python
> >>> linspace(0, 10, num=11)
> array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])
> ```

Sorry, I meant to ask : In which case does Numpy display the dtype ?

I'm adding the option in here so that it is now possible to disable the display of "dtype=". 
Currently, it is always displayed, so I set `ULAB_PRINT_DTYPE` to 1 by default, but my goal is to be able to deactivate it.

PR #604, opened 2023-04-24
Fix create arrange empty range

FelixNumworks:
Hi ! I found this bug when inputing `arange(0,0)`


v923z:
Many thanks for tracking it down! 

There seems to be an error in the expected file: https://github.com/v923z/micropython-ulab/actions/runs/4787405019/jobs/8514042245?pr=604

Finally, could you, please, increment the version number in https://github.com/v923z/micropython-ulab/blob/afc8e4e165bb2c90c9614776c04df51c30eb1145/code/ulab.c#L36? Since this is a minor bug fix, I would say, set it to 6.0.8.

v923z:
Thanks!

PR #616, opened 2023-05-12
add bitwise operators

v923z:
Implements the feature request https://github.com/v923z/micropython-ulab/issues/615

PR #617, opened 2023-05-15
ulab.numpy: implement sinc for creating audio filters

jepler:
This is useful for generating FIR filters using code snippets generated at https://fiiir.com/ (at least those with a rectangular window type; other window types need additional functions but we can revisit it later if needed)

I think this will come in handy for folks who are using the advanced features of our audio synthesizer module, synthio.

e.g., the following block now gives highly similar results on ulab or numpy:

```py
try:
    import numpy as np
except:
    from ulab import numpy as np

# Example code, computes the coefficients of a low-pass windowed-sinc filter.

# Configuration.
fS = 48000  # Sampling rate.
fL = 4000  # Cutoff frequency.
N = 23  # Filter length, must be odd.

# Compute sinc filter.
h = np.sinc(2 * fL / fS * (np.arange(N) - (N - 1) / 2))

# Normalize to get unity gain.
h /= np.sum(h)

# Applying the filter to a signal s can be as simple as writing
# s = np.convolve(s, h)

### DIFF
--- a/code/numpy/vector.c
+++ b/code/numpy/vector.c
@@ -570,6 +570,27 @@ MATH_FUN_1(sin, sin);
 MP_DEFINE_CONST_FUN_OBJ_1(vector_sin_obj, vector_sin);
 #endif
 
+#if ULAB_NUMPY_HAS_SINC
+//| def sin(a: _ArrayLike) -> ulab.numpy.ndarray:
+//|    """Computes the sine function"""
+//|    ...
+//|
+
+static mp_float_t ulab_sinc1(mp_float_t x) {
+    if (fpclassify(x) == FP_ZERO) {
+        return MICROPY_FLOAT_CONST(1.);
+    }
+    x *= MP_PI;
+    return MICROPY_FLOAT_C_FUN(sin)(x) / x;
+}
+
+static mp_obj_t vector_sinc(mp_obj_t x_obj) {
+    return vector_generic_vector(x_obj, ulab_sinc1);
+}
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: code/numpy/vector.c [hunk 1]
```diff
@@ -570,6 +570,27 @@ MATH_FUN_1(sin, sin);
 MP_DEFINE_CONST_FUN_OBJ_1(vector_sin_obj, vector_sin);
 #endif
 
+#if ULAB_NUMPY_HAS_SINC
+//| def sin(a: _ArrayLike) -> ulab.numpy.ndarray:
+//|    """Computes the sine function"""
```
v923z, line 575: There is a typo here, I believe.

v923z:
@jepler If you think that this is useful, then we should definitely add it. 

However, before merging, could you, please, add `build` to the requirements list, as you suggested in https://github.com/v923z/micropython-ulab/pull/616, so that the checks pass?

jepler:
This isn't critical, because it's quite possible to implement `sinc` in python (thanks @gamblor21):
```py
def sinc(x):
    y = np.pi * np.where(x == 0, 1.0e-20, x)
    return np.sin(y)/y
```

v923z:
> This isn't critical, because it's quite possible to implement `sinc` in python (thanks @gamblor21):

You should just merge it, if you are satisfied.
 
> ```python
> def sinc(x):
>     y = np.pi * np.where(x == 0, 1.0e-20, x)
>     return np.sin(y)/y
> ```

Indeed. In that case, it could've been implemented as a snippet.

PR #627, opened 2023-06-19
fix fetching submodules (cp build process change)

jepler:
In https://github.com/adafruit/circuitpython/pull/8070 we changed the commands for fetching submodules.

from what I can see, fetch-all-submodules is what fetch-submodules was; this is more than strictly necessary but should be the smallest change to get ulab building again.

This is trying to fix the underlying the cause of the build failure seen in #625


v923z:
@jepler Many thanks!

PR #629, opened 2023-06-22
Int overflow

FelixNumworks:
Fix multiple problems due to overflowing of int types + infinite arange

Also @v923z is there a way to contact you via email ? You can send me one at felix.richart@numworks.com ! Thanks a lot.


v923z:
Looks good, thanks!

PR #631, opened 2023-06-27
Fix -Wunused-variable diagnostic when !ULAB_SUPPORTS_COMPLEX

jepler:
`o_in` is only used in the SUPPORTS_COMPLEX case, so the variable definition needs to be moved accordingly.

### DIFF
--- a/code/ulab.c
+++ b/code/ulab.c
@@ -33,7 +33,7 @@
 #include "user/user.h"
 #include "utils/utils.h"
 
-#define ULAB_VERSION 6.3.3
+#define ULAB_VERSION 6.3.4
 #define xstr(s) str(s)
 #define str(s) #s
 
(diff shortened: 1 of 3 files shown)


v923z:
@jepler Indeed, many thanks for bringing this up! Could you, please, change the version number to 6.3.4, and then merge this?

code/ulab.c
```diff
@@ -33,7 +33,7 @@
 #include "user/user.h"
 #include "utils/utils.h"
 
-#define ULAB_VERSION 6.3.2
+#define ULAB_VERSION 6.3.3
```
v923z, line 36: Should be set to 6.3.4.

docs/ulab-change-log.md
```diff
@@ -1,5 +1,17 @@
 Fri, 12 May 2023
 
+version 6.3.3
```
v923z, line 3: 6.3.3 brings in an improvement on `polyval`.

v923z:
@jepler Many thanks, Jeff! I've merged this.

PR #650, opened 2023-12-08
Fix (u)int8 upcasting as per docs and numpy

### LINKED ISSUE #649: [BUG] Operations on uint8 don't respect upcasting rules
The documentation clearly states:

> ulab observes the following upcasting rules:
> 1. Operations on two ndarrays of the same dtype preserve their dtype, even when the results overflow.
> 2. ...

However the implementation does not respect this rule for uint8 types, which are upcast to uint16 regardless:

**To Reproduce**
```
>>> import ulab.numpy as np
>>> x = np.array([1, 2], dtype=np.uint8)
>>> x + x
array([2, 4], dtype=uint16)
```

**Expected behavior**
The result should have a uint8 dtype.

**Additional context**
Since this problem only regards the uint8 dtype and seems to be implemented rather consistently in this way it seems likely that this was an intentional choice at some point. However it is completely undocumented and I don't understand the rationale behind it.

The current behaviour means if the user wants to remain in "uint8-space", that is **not possible at all**, even using explicit casting. If the implementation were changed to match the docs however, users could explicitly cast one argument to uint16 to obtain the wanted result if needed, without breaking the other use case.

This is related to but separate from #396, which was partially solved by changing the 'downcasting' implementation for scalar arguments in #398. However adding a small integer constant to an u8 array still results in an u16 result due to this upcasting exception.

s-ol:
Close #649.

For testing with CircuitPython, see the branch https://github.com/s-ol/micropython-ulab/tree/fix-int8-upcasting.

Apologies if this breaks any tests that should have been updated, I couldn't get the test build to work on my system.


v923z:
The upcasting rules are given here: https://github.com/v923z/micropython-ulab/blob/4bde4efa9de1d002013c57805417aaef4da95740/code/ndarray_operators.c#L32, and here https://github.com/v923z/micropython-ulab/blob/4bde4efa9de1d002013c57805417aaef4da95740/code/ndarray_operators.c#L32.

code/ndarray_operators.c
```diff
@@ -181,11 +181,11 @@ mp_obj_t ndarray_binary_add(ndarray_obj_t *lhs, ndarray_obj_t *rhs,
...
     if(lhs->dtype == NDARRAY_UINT8) {
         if(rhs->dtype == NDARRAY_UINT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT16);
-            BINARY_LOOP(results, uint16_t, uint8_t, uint8_t, larray, lstrides, rarray, rstrides, +);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT8);
+            BINARY_LOOP(results, uint8_t, uint8_t, uint8_t, larray, lstrides, rarray, rstrides, +);
         } else if(rhs->dtype == NDARRAY_INT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT16);
-            BINARY_LOOP(results, int16_t, uint8_t, int8_t, larray, lstrides, rarray, rstrides, +);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT8);
```
v923z, line 187: This is wrong, because you're adding an `int8`, and a `uint8`, which should be upcast to `int16`.

code/ndarray_operators.c
```diff
@@ -264,11 +264,11 @@ mp_obj_t ndarray_binary_multiply(ndarray_obj_t *lhs, ndarray_obj_t *rhs,
...
     if(lhs->dtype == NDARRAY_UINT8) {
         if(rhs->dtype == NDARRAY_UINT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT16);
-            BINARY_LOOP(results, uint16_t, uint8_t, uint8_t, larray, lstrides, rarray, rstrides, *);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT8);
+            BINARY_LOOP(results, uint8_t, uint8_t, uint8_t, larray, lstrides, rarray, rstrides, *);
         } else if(rhs->dtype == NDARRAY_INT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT16);
-            BINARY_LOOP(results, int16_t, uint8_t, int8_t, larray, lstrides, rarray, rstrides, *);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT8);
```
v923z, line 270: Same here.

code/ndarray_operators.c
```diff
@@ -495,8 +495,8 @@ mp_obj_t ndarray_binary_subtract(ndarray_obj_t *lhs, ndarray_obj_t *rhs,
             results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT8);
             BINARY_LOOP(results, uint8_t, uint8_t, uint8_t, larray, lstrides, rarray, rstrides, -);
         } else if(rhs->dtype == NDARRAY_INT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT16);
-            BINARY_LOOP(results, int16_t, uint8_t, int8_t, larray, lstrides, rarray, rstrides, -);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT8);
```
v923z, line 498: And here.

code/ndarray_operators.c
```diff
@@ -509,8 +509,8 @@ mp_obj_t ndarray_binary_subtract(ndarray_obj_t *lhs, ndarray_obj_t *rhs,
         }
     } else if(lhs->dtype == NDARRAY_INT8) {
         if(rhs->dtype == NDARRAY_UINT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT16);
-            BINARY_LOOP(results, int16_t, int8_t, uint8_t, larray, lstrides, rarray, rstrides, -);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_INT8);
```
v923z, line 512: And here.

code/ndarray_operators.c
```diff
@@ -181,11 +181,11 @@ mp_obj_t ndarray_binary_add(ndarray_obj_t *lhs, ndarray_obj_t *rhs,
 
     if(lhs->dtype == NDARRAY_UINT8) {
         if(rhs->dtype == NDARRAY_UINT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT16);
-            BINARY_LOOP(results, uint16_t, uint8_t, uint8_t, larray, lstrides, rarray, rstrides, +);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT8);
```
v923z, line 184: This was clearly an error.

code/ndarray_operators.c
s-ol: Ah sorry, I got a bit carried away there between this and some other modifications I was trying.

v923z:
I think this is OK now, thanks! Could you, please, increment the sub-minor version here https://github.com/v923z/micropython-ulab/blob/4bde4efa9de1d002013c57805417aaef4da95740/code/ulab.c#L36, and add an entry here https://github.com/v923z/micropython-ulab/blob/master/docs/ulab-change-log.md? You'll also have to fix the test https://github.com/v923z/micropython-ulab/actions/runs/7153101407/job/19492990488?pr=650, but I think that's just a single line in https://github.com/v923z/micropython-ulab/blob/master/tests/2d/numpy/operators.py.exp, here https://github.com/v923z/micropython-ulab/blob/4bde4efa9de1d002013c57805417aaef4da95740/tests/2d/numpy/operators.py.exp#L97.

v923z:
OK, thanks!

PR #651, opened 2023-12-12
Implement integer array indexing

### LINKED ISSUE #607: [FEATURE REQUEST] add indexing via array
I would like the ability to use one ndarray to index into another one, aka “Advanced indexing” (https://numpy.org/doc/stable/user/basics.indexing.html)

eg:

x = np.arange(10, 1, -1)
-> x
array([10,  9,  8,  7,  6,  5,  4,  3,  2])
-> x[np.array([3, 3, 1, 8])]
array([7, 7, 9, 2])

s-ol:
Close #607.

This implements integer array indexing with the following limitations as of now:

TODO:
- [x] u/int index types
- [x] multi-dimensional index arrays
- [ ] negative indices
- [ ] index array range-checking
- [ ] slice-assignment

non-goals for this PR:
- support N-dim target types
- support float index types (debatable)

### DIFF
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -1197,6 +1197,94 @@ static mp_obj_t ndarray_from_boolean_index(ndarray_obj_t *ndarray, ndarray_obj_t
     return MP_OBJ_FROM_PTR(results);
 }
 
+
+static mp_obj_t ndarray_from_integer_index(ndarray_obj_t *ndarray, ndarray_obj_t *index) {
+    if(ndarray->ndim > 1) {
+        mp_raise_ValueError(MP_ERROR_TEXT("only supports 1-dim target arrays"));
+    }
+
+    if(!ndarray_is_dense(ndarray)) {
+        mp_raise_ValueError(MP_ERROR_TEXT("only supports dense target arrays"));
+    }
+
+    // TODO: range-check index values against ndarray->shape[ULAB_MAX_DIMS-1]
+    // TODO: normalize or handle negative indices in loop (without modifying index)
+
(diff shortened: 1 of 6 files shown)


s-ol:
Note: when compiling this from latest CircuitPython master, I need the following patch to avoid a GCC false positive:

```patch
diff --git a/code/ndarray.c b/code/ndarray.c
index c41c01a..46c84b3 100644
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -496,7 +496,10 @@ bool ndarray_is_dense(ndarray_obj_t *ndarray) {
     // the array should be dense, if the very first stride can be calculated from shape
     int32_t stride = ndarray->itemsize;
     for(uint8_t i = ULAB_MAX_DIMS - 1; i > ULAB_MAX_DIMS-ndarray->ndim; i--) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
         stride *= ndarray->shape[i];
+#pragma GCC diagnostic pop
     }
     return stride == ndarray->strides[ULAB_MAX_DIMS-ndarray->ndim] ? true : false;
 }
```

v923z:
You seem to bail out, when the array is not dense, but I think your `INDEX_LOOP` actually handles the general case.

v923z:
I think it's OK to not deal with `float` indices. Also, make this optional by defining a variable in `ulab.h`.

v923z:
If I look at your `INDEX_LOOP`, I have the feeling that you can save half of the space, if you copy the contents of the RAM instead of retrieving the value. You could treat `int8/uint8`, and `int16/uint16` on the same footing, because only the size matter, but not the signedness of the value.

PR #653, opened 2023-12-25
fix the `np.delete` bug

hiltay:
fix the `np.delete` bug in simple way.
now,if the second positional argument is empty list `[]`,return the original array instead.


v923z:
Thanks! Could you, please, increment the version number here https://github.com/v923z/micropython-ulab/blob/e32920645c8c5e6153f29e4e33ca7e3b7087aef1/code/ulab.c#L36, and add an entry here https://github.com/v923z/micropython-ulab/blob/master/docs/ulab-change-log.md? 

Also, it would be great, if this error was addressed in https://github.com/v923z/micropython-ulab/blob/master/tests/2d/numpy/delete.py.

v923z:
When you modify the unittest, you have to update the expected file: https://github.com/v923z/micropython-ulab/blob/master/tests/2d/numpy/delete.py.exp. 

https://github.com/v923z/micropython-ulab/#testing

hiltay:
I'm very willing to contribute and test:)
anything else to add?

v923z:
> I'm very willing to contribute and test:) anything else to add?

I think this is complete now, and as soon as the workflow finishes, I merge the PR. Many thanks for contributing!

Issue #662, opened 2024-02-20
[BUG] Some type annotations don't include scalars


kbsriram:
**Describe the bug**
Ulab version: `6.5.1-2D-c`

Type annotations on some functions don't appear to include scalars.

**To Reproduce**

For instance, this code:

```python
from ulab import numpy as np

np.acos(0)
```
when used with the [current type annotation](https://github.com/v923z/micropython-ulab/blob/acfec3e9af5d351df88a7d71786f11a12f7545d3/code/numpy/vector.c#L237-L243) extracted by circuitpython produces this error with `mypy`.
```shell
$ export MYPYPATH=${HOME}/circuitpython/circuitpython-stubs
$ mypy --disallow-untyped-defs type_test.py
type_test.py:3: error: Argument 1 to "acos" has incompatible type "int"; expected "Union[ndarray, List[float], Tuple[float], range]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

**Expected behavior**
The code accepts both scalars and `_ArrayLike` - so I was expecting type-checking to work.

**Additional context**
Would one option be to add a TypeAlias that unions scalars with _ArrayLike, and use that for methods that accept or produce both scalars and arrays?

v923z:
The issue concerns the documentation and not the code directly, right?

v923z:
OK. Then it's not a bug.

PR #663, opened 2024-02-28
Update type annotations in compare.c and vector.c

kbsriram:
- Add type annotations for functions in compare.c
- Update annotations in vector.c to match behavior

This should address at least the specific issues noted in https://github.com/adafruit/circuitpython/issues/8715 and https://github.com/v923z/micropython-ulab/issues/662 . (As I'd filed the original issue here for a more localized docubug :smile:) please of course feel free to re-purpose the issue, or close it to start a new generic tracking issue later as seems appropriate.


v923z:
Many thanks for your contribution!

PR #664, opened 2024-02-29
Remove the STATIC macro.

Gadgetoid:
Trying to get ahead of the changes proposed in micropython/micropython#13763.

I believe this change can be made safely ahead of MicroPython merging that PR, since `STATIC` should always resolve to `static` in production.

Methodology:

* `find -name *.c  | xargs sed -Ei "s/(^| )STATIC($| )/\1static\2/"` to automate replacement in .c files
* `grep -r STATIC` and manual fixes to the docs

### DIFF
--- a/code/ndarray.c
+++ b/code/ndarray.c
@@ -917,7 +917,7 @@ ndarray_obj_t *ndarray_from_iterable(mp_obj_t obj, uint8_t dtype) {
     return ndarray;
 }
 
-STATIC uint8_t ndarray_init_helper(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+static uint8_t ndarray_init_helper(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
     static const mp_arg_t allowed_args[] = {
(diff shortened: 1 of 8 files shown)


jimmo:
> I believe this change can be made safely ahead of MicroPython merging that PR

Yes, definitely. Good idea @Gadgetoid !

v923z:
Many thanks for keeping an eye on this!

Gadgetoid:
Note: This needs a new compatibility release, since MicroPython have now shipped v1.23.0 with the STATIC macro removed - https://github.com/micropython/micropython/releases/tag/v1.23.0

v923z:
@Gadgetoid I think I'm a bit confused now: the PR addressed the issue of the static keyword ahead of time, so we should be in sync with 1.23, shouldn't we?

Gadgetoid:
> Or do you just simply want to create a release to indicate this fact?

This! I think it's worth tagging a release though as you point out it's not at all required.

In hindsight I think I've just been submoduling and building against arbitrary points in the master branch anyway 😬

v923z:
I think you bring up a valid point. So, here you go: https://github.com/v923z/micropython-ulab/releases.

Gadgetoid:
Thank you!

PR #669, opened 2024-06-03
CI: Update some deprecated dependency versions

Gadgetoid:
Notably bumps Python to prevent macOS builds failing, and bumps "checkout" and "setup-python" versions to avoid deprecation warnings.

Tests/builds still seem to be failing so I've also enabled "continue-on-error" to try and make it run *all* the tests - for a better picture of what's failing and where- but seem to have failed to make this work across both MicroPython and CircuitPython tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,10 +28,10 @@ jobs:
       env:
         GITHUB_CONTEXT: ${{ toJson(github) }}
       run: echo "$GITHUB_CONTEXT"
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v1
+    - name: Set up Python 3.12
+      uses: actions/setup-python@v5
       with:
-        python-version: "3.10"
+        python-version: "3.12"
 
     - name: Install requirements
       run: |
@@ -44,10 +44,10 @@ jobs:
         gcc --version
         python3 --version
     - name: Checkout ulab
-      uses: actions/checkout@v1
+      uses: actions/checkout@v4
 
     - name: Checkout micropython repo
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
       with:
         repository: micropython/micropython
         path: micropython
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I believe the CircuitPython builds failing is due to an oversight in how the custom config supplied via `MP_CONFIGFILE` is handled. `shared/timeutils/timeutils.h` imports `mpconfigport.h` explicitly, causing some duplication errors between the default config and the custom config.

Gadgetoid:
Okay after some helpful feedback from tannewt, it looks like `MP_CONFIGFILE` is a vestigial non-feature that cannot be relied upon in CircuitPython.

The good news, however, is that aiui `
sed -e '/MICROPY_PY_UHASHLIB/s/1/0/' < circuitpython/ports/unix/mpconfigport.h > circuitpython/ports/unix/mpconfigport_ulab.h` did nothing, because `MICROPY_PY_UHASHLIB` does not appear in `unix/mpconfigport.h`. Adding a commit on to test this assumption seems to get CircuitPython CI up and running again.

The last hurdle:

```
2 tests failed: cholesky logspace
\nFAILURE cholesky.py
--- cholesky.py.exp	2024-06-03 22:44:06
+++ cholesky.py.out	2024-06-03 22:44:06
@@ -5,5 +5,5 @@
        [-1.0, 1.0, 3.0]], dtype=float64)
 array([[4.242640687119285, 0.0, 0.0, 0.0],
        [5.185449728701349, 6.565905201197403, 0.0, 0.0],
-       [12.72792206135786, 3.046038495400855, 1.649742247909068, 0.0],
-       [9.899494936611665, 1.624553864213789, 1.849711005231386, 1.392621247645583]], dtype=float64)
+       [12.72792206135786, 3.046038495400855, 1.649742247909066, 0.0],
+       [9.899494936611665, 1.624553864213789, 1.849711005231389, 1.392621247645579]], dtype=float64)
```

I ... am not qualified to know what might have caused this difference in these specific cases. Perhaps the laws of natural physics shifted slightly, the wind direction changed, someone patched a CPU vulnerability or some mysterious and wonderful nonsense affected floating point precision enough to throw these out of whack. Possibly just a fix-the-tests issue?

v923z:
@Gadgetoid Philip, many thanks for getting to the bottom of this! I would just let it pass for now, and I can fix the single test case that's causing the problem.

PR #674, opened 2024-07-16
SVD unittests

EmDash00:


EmDash00:
Getting an odd bug on one of my unittests where one of the prints causes them to crash. 

https://github.com/EmDash00/micropython-ulab/blob/5587d4e69dd7e635d5e6223aabc7019057631d2d/tests/2d/scipy/svd.py#L94

```py
A = rng.random((15, 10))
U, S, Vh = sp.linalg.svd(A)
# print(np.dot(U * S, Vh))
B = np.dot(np.dot(U, S), Vh)
print("15x10 Random Compare:", np.max(abs(A - B)) < tol)
print()
```

I'm not sure why it doesn't like the print line. All the other prints in the unittest seem fine.

v923z:
OK, thanks, I'll take a look.

PR #675, opened 2024-07-17
Removed Pycom reference

andypiper:
UInfortunately Pycom is no longer a thing, removed from the platforms list.


v923z:
Many thanks for the correction!

PR #678, opened 2024-08-06
Correct `"a[a < 5]"` in ulab-ndarray.ipynb

page200:
The value of `a[a < 5]` should be referred to as `"a[a < 5]"`, not `"a < 5"`, because `a < 5` has a different value.


v923z:
Indeed. Thanks for pointing this out!

PR #680, opened 2024-08-23
Add missing type hints to `carray`

elpekenin:
Very tiny PR :)

Noticed this while looking though CircuitPython's typing issues.

Might as well hint `real()` and `imag()` with `builtins.complex` too, but since it seems not to be supported yet, decided to only hint the supported types even though it contradicts the docstrings.

### DIFF
--- a/code/numpy/carray/carray.c
+++ b/code/numpy/carray/carray.c
@@ -25,9 +25,11 @@
 
 #if ULAB_SUPPORTS_COMPLEX
 
+//| import builtins
+//|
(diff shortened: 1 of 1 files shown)


elpekenin:
`T = TypeVar("T", bound=Union[builtins.complex, ulab.numpy.ndarray])` to indicate that the output is the same type as input would be slightly better, i can change it if prefered, but thought it would be best to keep things simple

v923z:
Thanks!

PR #684, opened 2024-09-08
numpy/random.c: fix use of MICROPY_PY_RANDOM_SEED_INIT_FUNC

dhalbert:
In `numpy/random.c`:

The preprocessor guard for `MICROPY_PY_RANDOM_SEED_INIT_FUNC` was not complete: the value was used even if it was undefined.

There were also inconsequential trailing spaces and a trailing newline that were cleaned up by emacs when I edited the file.

### DIFF
--- a/code/numpy/random/random.c
+++ b/code/numpy/random/random.c
@@ -41,7 +41,7 @@ MP_DEFINE_CONST_OBJ_TYPE(
     MP_QSTR_generator,
     MP_TYPE_FLAG_NONE,
     print, random_generator_print,
-    make_new, random_generator_make_new, 
+    make_new, random_generator_make_new,
     locals_dict, &random_generator_locals_dict
 );
 #else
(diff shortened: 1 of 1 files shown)


dhalbert:
FYI, we're merging micropython v1.23 into circuitpython, and this was a blocker (though we may need to turn off some functionality like random if it's not fitting on some boards).

dhalbert:
I don't know what to make of the build failures, but I think they are unrelated.

v923z:
> I don't know what to make of the build failures, but I think they are unrelated.

Yeah, that's totally unrelated, I'd just let it pass for now. I've meant to fix that, but I forgot. Thanks for bringing it up!

v923z:
> In `numpy/random.c`:
> 
> The preprocessor guard for `MICROPY_PY_RANDOM_SEED_INIT_FUNC` was not complete: the value was used even if it was undefined.
> 
> There were also inconsequential trailing spaces and a trailing newline that were cleaned up by emacs when I edited the file.

Many thanks!

PR #695, opened 2024-11-25
Overlap and Add algorithm for fast convolution of long array

eleroy:
Hi everyone,

This an implementation of the oaconvolve method that may be faster than convolve espiecielly when dealing with long arrays. 

Few things I'm not sure:
- I don't think this will work without complex support and I'm not sure how to change the code to do so
- I'm not sure there si a need for m_free at the end of the method
- I just realized there are micropython contributing guidelines that I probably did not completely follow

The method takes the same arguments as the convolve method.

### DIFF
--- a/code/numpy/filter.c
+++ b/code/numpy/filter.c
@@ -10,7 +10,7 @@
  *               2020 Scott Shawcroft for Adafruit Industries
  *               2020-2021 Zoltán Vörös
  *               2020 Taku Fukada
-*/
+ */
 
 #include <math.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


eleroy:
Pff...sorry i forgot to move the method to signal, left it where i built it in filter.

I also saw it didn't built since i added the m_free that was not well used (IS it useful to free the buffer at the end or it will be free anyway when the method returns?)

I'll modify the PR to address your remarks.

v923z:
And I think that the problem of complexes has to be solved. It wouldn't make sense to include complex support for the sake of a single method.

v923z:
The checks fail here: https://github.com/v923z/micropython-ulab/actions/runs/12020316249/job/33509066852?pr=695#step:8:317.

v923z:
> I also saw it didn't built since i added the m_free that was not well used (IS it useful to free the buffer at the end or it will be free anyway when the method returns?)

In this case, it shouldn't really matter.

eleroy:
added:
- Support without complex number
- Documentation for oaconvolve
- Bump ulab version
- Test for oaconvolve (same as convolve but with reduced absolute tolerance, it seems the method is not as accurate as convolve)

docs/manual/source/scipy-signal.rst
```diff
@@ -2,9 +2,10 @@
 scipy.signal
```
v923z, line 2: The documentation is actually generated from this: https://github.com/v923z/micropython-ulab/blob/master/docs/scipy-signal.ipynb. The jupyter notebook allows you to run `micropython` code directly, so you don't have to copy anything.

code/scipy/signal/signal.c
```diff
@@ -120,11 +122,180 @@ mp_obj_t signal_sosfilt(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_ar
 MP_DEFINE_CONST_FUN_OBJ_KW(signal_sosfilt_obj, 2, signal_sosfilt);
 #endif /* ULAB_SCIPY_SIGNAL_HAS_SOSFILT */
 
+#if ULAB_SCIPY_SIGNAL_HAS_OACONVOLVE
+
+mp_obj_t signal_oaconvolve(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args)
+{
```
v923z, line 128: This doesn't conform to `micropython`'s coding standards: https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md#c-code-conventions
`Put opening braces on the end of the line it belongs to, not on a new line.`

code/scipy/signal/signal.c
```diff
@@ -19,6 +19,8 @@
 #include "../../ulab.h"
 #include "../../ndarray.h"
 #include "../../numpy/carray/carray_tools.h"
+#include "../../numpy/fft/fft_tools.h" // For fft
```
v923z, line 22: It's clear that it's for the FFT, you can remove the commetn;-)

code/scipy/signal/signal.c
```diff
@@ -19,6 +19,8 @@
 #include "../../ulab.h"
```
v923z, line 19: Add your name/year at the beginning of the file!

v923z:
> added:
> 
>     * Support without complex number
> 
>     * Documentation for oaconvolve
> 
>     * Bump ulab version
> 
>     * Test for oaconvolve (same as convolve but with reduced absolute tolerance, it seems the method is not as accurate as convolve)

Many thanks! I think it's almost ready to go. I added a couple of small comments. If those can be resolved, this could be merged.

docs/manual/source/scipy-signal.rst
eleroy: Ha ok, I made the changes in the jupyter notebook but I'm not sure how to run it (my system is windows).

PR #701, opened 2024-12-30
Add  keepdims keyword argument

v923z:
Add requested keyword argument keepdims to some numerical functions


v923z:
@kwagyeman I think this should address the issue that you raised. I'll re-work this in the coming days, for the simple reason that others have requested the `out` keyword argument, but that's not going to change anything with respect to `keepdims`. In other words, you won't even notice that, except, perhaps for the code size.

kwagyeman:
Awesome! Once this is merged we can integrate it into OpenMV.

v923z:
> Awesome! Once this is merged we can integrate it into OpenMV.

If this is urgent for you, I can merge this right away. I really see no reason for waiting otherwise.

kwagyeman:
It's not urgent. Please test and build whatever confidence you need in the PR along with adding the other feature.

kwagyeman:
@v923z - Any updates? Back to work now and would like to get this merged.

v923z:
I can merge the current implementation right away. I have the skeleton of the improved version, but I can't promise at the moment that I can flesh out all detail this weekend. So, if this is urgent, I would propose that we merge what we have now, and future improvements won't affect the user, for the interface will remain the same.

v923z:
I would emphasize that this is a correctly working implementation.

kwagyeman:
That sounds good. That would unblock me after you release what you have.

kwagyeman:
I can wait for that to be updated.

v923z:
OK. I can definitely fix that during the weekend.

PR #746, opened 2026-02-07
meshgrid (#740)

### LINKED ISSUE #740: [FEATURE REQUEST] meshgrid
When working through numpy code for AI ML with chatgpt it often wants to use https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html.

Sparse and Copy could be left to their defaults and not supported.

iyassou:
Hello 👋

Long time lurker first time contributor, love this repo ❤️

I've added an implementation of `np.meshgrid` which closes #740. It's meant to be equivalent to NumPy's version with `copy=True` and `sparse=False` (default kwargs).

I've added simple test cases and updated the relevant docs.

### DIFF
--- a/code/ulab.c
+++ b/code/ulab.c
@@ -33,7 +33,7 @@
 #include "user/user.h"
 #include "utils/utils.h"
 
-#define ULAB_VERSION 6.11.0
+#define ULAB_VERSION 6.12.0
 #define xstr(s) str(s)
 #define str(s) #s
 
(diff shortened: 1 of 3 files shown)


v923z:
> I've added an implementation of `np.meshgrid` which closes #740. It's meant to be equivalent to NumPy's version with `copy=True` and `sparse=False` (default kwargs).

This looks good, many thanks for investing the time in implementing `meshgrid`. I'd like to ask you to change the version string to 6.12 here: https://github.com/v923z/micropython-ulab/blob/a8b25eff3eea742b949012e512de7c69886e7794/code/ulab.c#L36

Otherwise, this PR is good to go.

v923z:
@iyassou Thanks again for your efforts!

PR #749, opened 2026-04-08
ndarray: Fix four bugs in the modulo operator (% and %=)

kwagyeman:
The recently added modulo operator had four bugs, two of which caused
memory corruption or silent wrong results at runtime.

### Binary `%` — `uint16 % uint8` result dtype

`ndarray_binary_modulo` allocated the result array as `NDARRAY_UINT8`
but then looped with `uint16_t` arithmetic via `BINARY_LOOP`. Because
`BINARY_LOOP` writes through a `uint16_t` pointer but the array's
`itemsize` is 1 byte, this caused out-of-bounds stores that corrupted
both the result values and adjacent memory. Fixed by allocating the
result as `NDARRAY_UINT16` to match the left operand.

### Inplace `%=` — four bugs

1. **Feature guard typo**: `ulab.h` defined `NDARRAY_HAS_INPLACE_MODU`
   (missing `LO`), so `NDARRAY_HAS_INPLACE_MODULO` was never set and
   `ndarray_inplace_modulo` was compiled out entirely. All `%=`
   operations silently fell back to binary `%` + rebind.

2. **Missing variable declarations**: Every `INLINE_MODULO_FLOAT_LOOP`
   call referenced `larray` and `rarray` but they were never declared —
   a compile error waiting to happen once the typo above was fixed.

3. **Duplicate dtype branch**: The second `else if` checked
   `NDARRAY_UINT8` again instead of `NDARRAY_INT8`, making
   `float %= int8` a silent no-op.

4. **Integer lhs unhandled**: Only `NDARRAY_FLOAT` lhs was dispatched;
   all integer lhs types fell through and returned the array unchanged.
   Added full integer dispatch using `INPLACE_LOOP` with `%=`. Integer
   lhs `%=` float rhs now raises `TypeError`, matching NumPy's in-place
   casting semantics.

## Test plan

- Updated `tests/2d/numpy/modulo.py` and `modulo.py.exp` to reflect
  correct dtypes and in-place semantics, including expected `TypeError`
  for integer `%=` float
- Added a regression section with one targeted case per bug

### DIFF
--- a/code/ndarray_operators.c
+++ b/code/ndarray_operators.c
@@ -292,7 +292,7 @@ mp_obj_t ndarray_binary_modulo(ndarray_obj_t *lhs, ndarray_obj_t *rhs,
         }
     } else if(lhs->dtype == NDARRAY_UINT16) {
         if(rhs->dtype == NDARRAY_UINT8) {
-            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT8);
+            results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_UINT16);
             BINARY_LOOP(results, uint16_t, uint16_t, uint8_t, larray, lstrides, rarray, rstrides, %);
         } else if(rhs->dtype == NDARRAY_INT8) {
             results = ndarray_new_dense_ndarray(ndim, shape, NDARRAY_FLOAT);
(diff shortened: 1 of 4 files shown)


v923z:
@kwagyeman Yes, it's good, many thanks for putting in the effort!

v923z:
@kwagyeman Here is the release tag: https://github.com/v923z/micropython-ulab/releases/tag/6.12.

kwagyeman:
Thanks!

PR #751, opened 2026-07-22
Fix ULAB_SCIPY_HAS_*_MODULE flags

dhalbert:
Except for `linalg.c`, the individual modules in `scipy` did not have their contents guarded by the appropriate `ULAB_SCIPY_HAS_*_MODULE` flags. 

- Add the appropriate `#if` guards around the code contents in several `scipy` modules.
- Fix a missing comma in a Python function signature in `integrate.c`.
- Remove trailing blanks in several files. (My editor did this automatically.) A pre-commit action could catch this in the future.

This came up because I was updating `ulab` to 6.12.1 in CircuitPython. But some builds no longer fit, so I had to turn off some features. The new `scipy.integrate` features could be turned off individually, but not the whole `integrate` module. In addition, some other `scipy` modules could not be turned off completely.

An LLM under my direct supervision did the mechanical work of looking for missing guards and making most of the changes. I touched up the results.

### DIFF
--- a/code/scipy/integrate/integrate.c
+++ b/code/scipy/integrate/integrate.c
@@ -9,18 +9,18 @@
  *
  * References:
  * - Dr. Robert van Engelen, Improving the mp_float_t Exponential Quadrature Tanh-Sinh, Sinh-Sinh and Exp-Sinh Formulas,
- *   2021, https://www.genivia.com/qthsh.html 
+ *   2021, https://www.genivia.com/qthsh.html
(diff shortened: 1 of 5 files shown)


dhalbert:
I am puzzled by the micropython build failures here. Some are due to slightly different floating-point printed values, and some are due to what looks like some `#include` issues. Maybe there are changes upstream in `micropython`.

v923z:
Thanks for catching this!

PR #752, opened 2026-07-22
update actions to use Node 24

dhalbert:
GitHub CI was raising warnings about Node 20 actions. Update actions to Node 24.

I updated to `@v6` actions. `checkout@v7` and `setup-version@v7` are available but they are very new. I'd wait a while to let them shake out.


dhalbert:
Similar build failures as #751, so I think this is due to upstream changes in micropython.

PR #755, opened 2026-08-06
update tests/1d/complex results due to slight changes in float printing

### LINKED ISSUE #756: [BUG] not all the tests run during CI
Version 6.12.1+ (https://github.com/v923z/micropython-ulab/commit/01ad8a5fff89d7aeb68ef1827d1ba8f83f6462a7)

I was trying to fix the recent CI build failures, which were due to slightly different floating point prints. So I did PR #755.

I initially got wrong which `.exp` files to change. There were leftover `.out` and `.exp` files in the top-level directory after running `./build.sh`, but I missed which set of tests they corresponded to. The force-push in #755 fixed that.

However, an LLM noticed that a number of tests are being skipped, apparently due to a `break` instead of a `continue` in `test-common.sh`

```sh
for level2 in numpy scipy utils complex; do
    rm -f *.exp
    if [ ! -d tests/"$level1"/"$level2" ]; then
        break;          # exits the entire level2 loop, not just this iteration
    fi
```

Only `tests/2d` has all four subdirectories `complex`, `numpy`, `scipy`, and `utils.
`1d`, `3d`, and `4d` contain only`complex` and `numpy`:

```
1d: complex numpy
2d: complex numpy scipy utils
3d: complex numpy
4d: complex numpy
```

So when `test-common.sh` runs, the loop exits prematurely due to the `break`:

```
1d/numpy   runs
1d/scipy   MISSING -> break, rest of 1d abandoned   (1d/complex never runs)
2d/numpy   runs
2d/scipy   runs
2d/utils   runs
2d/complex runs
3d/numpy   runs
3d/scipy   MISSING -> break, rest of 3d abandoned   (3d/complex never runs)
4d/numpy   runs
4d/scipy   MISSING -> break, rest of 4d abandoned   (4d/complex never runs)
```

So `tests/2d/complex` is the only `complex` test directory that ever executes

After this is fixed, there will probably new test failures, due to at least the same reasons that are fixed in #755.

A related problem: `rm -f *.exp` at line 8 clears stale `.exp` files before each run, but `.out` files are never cleaned up and accumulate in the repository root for the whole script run. Adding `.out` to that `rm` would make that cleaner.

This issue was originally written by an LLM, but I cleaned it up a lot for clarity.

dhalbert:
micropython builds are failing due to slightly different test results from newer versions of micropython. Update the test results.

Also fix #756


v923z:
Thanks a lot for taking a look at this! I believe, the proper way of solving this would be using `isclose`, because otherwise, a test might pass in a windows or linux environment, but then fail on a macintosh, or vice versa.

dhalbert:
I'm going to close this in favor of an upcoming PR with `isclose()`.

